PR #2, opened 2018-04-28
Lesson01

ghassan24601:
Is `test_intsum2()` wrong ?
There is a function call for ` g = gen.intsum2() ` and ther exercise doesn't say create a second intsum function.


ldconejo:
Hi Ghassan,

You're right. test_intsum2() is there in error. Thanks for point it out.

-Luis

PR #15, opened 2018-06-02
Lesson05|Debug recursive function

nDruP:
Used pdb on recursive.py to find the logic issue with my_fun.
Debug log is recursive_debug_log.txt


ldconejo:
Proposed explanation is correct.

PR #21, opened 2018-06-09
lesson 1 files

kuhnbt7:


ldconejo:
Hi Ben,
Make sure you run tests with _test_generator.py_, provided as part of the assignment (disable the _test_intsum2_ method, as that one is not required), make sure it passes and include it on your pull request. This will require you to modify the name of some of your methods or to modify the method names in the unit tests to match the ones you used.

kuhnbt7:
Oops I totally missed that, sorry about that. I'll modify and resubmit.

kuhnbt7:
Hi Luis, I modified the generator code and all the tests pass now. I'm
trying to make a new pull request and it just takes me to the old pull
request. I'm not super experienced with git - I think my new commit was
just added to the existing pull request? Can you let me know if you can see
the modified file?

Regarding the iterator file, I think I implemented it to take a start,
stop, and step parameter...the last step of the instructions says to make
it match range() so I made some modifications. If you give range one
argument (range(10)), then the first argument acts as the stop parameter.
But if you give it 2 parameters (range(5,10)), then the first argument is
the start parameter and the second is the stop parameter. So that's what I
changed my iterator to do. I made a comment in my code asking if there's a
simpler way than the way I did it, but I think the code I wrote is
functional. Let me know if there are problems that I overlooked.

Thanks,
Ben

ldconejo:
Hi Ben,

The generator tests are now passing, good job! For the iterator, I failed to see at first that you were using Truthiness to essentially use the same variables as boolean and integers, which is quite interesting. It might be confusing, but it's still valid and you actually went the extra distance in having a single iterator class that pretty much does everything :-)

There's one more issue which I'd like to take a quick look at, get it fixed and move on to the next assignment: If you try to run with IterateMe_1(2,7,2), you would expect the code to return something like this:

2
4
6

Alas, it still returns the following:

0
1
2
3
4

The problem is the code you have on line 34.

kuhnbt7:
Hi Luis, I can't seem to replicate the problem you're referring to. When I
run IterateMe_1(2, 7, 2) I get the result you expect (2, 4, 6). I'm running
the same version of the code that I pushed to github. Can you clarify?

Thanks,
Ben

ldconejo:
That's interesting. The reason it should not work correctly is that you're calling _self.__init__()_ from ___iter__(self)_ method (that's what I see on line 34), which reinitializes the class with default values (start=False, stop=False, step=False), which causes you to get the default iteration sequence (0, 1, 2, 3, 4).

-Luis

PR #27, opened 2018-06-11
lesson02 generators_closures2.py

kmsnyde:
Created a generator and closure for lesson02 in same file.


ldconejo:
Hi Karl,

Works as expected. One bit of feedback in the generator one is to have the generator yield all matches instead of just two. You can actually achieve that by doing the same you do to print all results in _high_e_sort_. I really liked your implementation of closures, letting the returned function still take another parameter (col).

PR #30, opened 2018-06-13
Lesson03 context manager and recursive function

kmsnyde:


ldconejo:
Works as expected. Here's a different way to show the exception message as part of the _exit_ method of the class, which yields a more elegant output (this example comes from another student's solution):

        if exc_type:
            print(exc_val)
        return True

PR #40, opened 2018-06-19
Lesson05 | Logging

nDruP:
adjusted simple.py to work with syslogserver.


ldconejo:
Hi Andrew,

Keep in mind that errors need to be logged to syslogserver. Try setting syslogserver.py to use the local host (IP: 127.0.0.1, port 514 is fine) and also get server_handler to point to that address / port, so that you can try the whole thing on the same machine. When you launch simple.py (with syslogserver.py running in another terminal/command window) you will see a message like this on syslogserver:

127.0.0.1 :  <11>simple.py:41  ERROR Tried to divide by zero. Var i was 50. Recovered gracefully.

Let me know if you have any questions!

PR #41, opened 2018-06-19
Lesson02 Assignment - Generators/Closures

csdotson:
**Notes**: Completed lesson 2 assignment. I *think* I had the right idea on the closure part of the assignment, but let me know if it's not quite what you had in mind. Thanks!


ldconejo:
Code works correctly. I used the following lines of code to test high_energy_closure.py:

test = energy_track_finder(0.8)
for track in test():
    print(track)

PR #47, opened 2018-06-22
lesson05

kmsnyde:


ldconejo:
Hi Karl,
For this one, the way it should work is that, while running syslogserver.py in a separate terminal and then launching simple.py, syslogserver should report the remote error. Something like this:

`127.0.0.1 :  <11>simple.py:47  ERROR Tried to divide by zero. Var i was 50. Recovered gracefully`

Two things to consider in your simple.py code: 

1. Make sure you're setting up your _syslog_handler_ variable to use the IP address that _syslogserver.py_ is using.
2. You're creating _syslog_handler_ but are you adding that handler to the logger?

Thanks,

Luis

kmsnyde:
Updated various files.

Added DatagramHandler to work in windows.  Tested successfully from command line for syslogserver.py and simple.py.

ldconejo:
Looks good now, good job!

PR #49, opened 2018-06-22
Lesson03 Assignment - Context Manager/Recursive Factorial

csdotson:
**Notes**: Complete lesson 3 assignment - context manager and recursive solution for the factorial function. I had some trouble on the context manager; let me know if what I have isn't quite correct. Thanks!


ldconejo:
Works great, one thing you could try is to add a _return True_ to your __exit__ method, so that it exits gracefully. It will look like this:

```
    def __exit__(self, e_type, e_val, e_tb):
        if e_type is ValueError:
            print('Warning: Too many boats for the locke!')
        **return True**
```

PR #52, opened 2018-06-25
Lesson01 - Comprehensions

DeFeNdog:


ldconejo:
Hi Carlos,

On iterator_1.py, when you break the first loop and then resume, it should continue the iteration, but it's instead restarts it. Take a look at that and and let me know when it's fixed. The problem is with the __iter__ method.

ldconejo:
Hi Carlos,

Please disregard my previous comment. I realize that the assignment does ask you to make your iterator identical to range, which implies resetting the sequence.

Thanks,

Luis

PR #53, opened 2018-06-25
Lesson07|mailroom

nDruP:
donor_peewee_model creates a table and populates it with data.
added a class method to donor_dict that parses the Donation table to initialize from donor.db
added a function to mailroom that saves new gifts to donor.db

donor_db.py was an attempt to refactor donor_dict to query from donor.db for info, but I think it might be less efficient than parsing the data so I gave up.


ldconejo:
Hi Andrew,

Could you add code to enable deleting/updating entries in the database? The idea is that the user would be able to change the amount of donation or even delete one altogether (or maybe delete all donations by a user). I will verify deletion/update using a DB browser.

Thanks,

Luis

nDruP:
I completely forgot about that part of the assignment....
I've updated mailroom.py functionality and fleshed out donor_db.py

ldconejo:
Thanks, Andrew. Looks good :-)

PR #55, opened 2018-06-25
Finished both assignments.  

hortw1991:


hortw1991:
While driving home, I just realized I'm dumb or under caffeinated and used a list comprehension again.  Fixing this now.

PR #66, opened 2018-07-01
Lesson01 Added

natrod22:
I talked with GitHub and they helped me fix my commit/push issue.


ldconejo:
Hi Natalie,

Great news that you're now able to commit your changes. On your pull request, make sure that you're able to run (an pass) testing for your code. Currently, there's an error in the import you use for _test_generator.py_ which causes pytest to fail when I run it. Let me know if you have any questions!

Thanks,

Luis

natrod22:
I updated my test_generator.py file. Everything should work now :)

ldconejo:
Hi Natalie,

Almost there :-) Try running the following command on the same directory where _test_generator.py_ resides (install pytest if you don't have it already):

python -m pytest 

The output should look like this:

```
    def test_primes():
        g = gen.primes()
        for val in [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97]:
>           assert next(g) == val
E           assert 1 == 2
E            +  where 1 = next(<generator object primes at 0x105c25c78>)

test_generator.py:49: AssertionError
============================================================================== 1 failed, 5 passed in 0.08 seconds ==============================================================================
ldconejo-mac08:Lesson01 ldconejo$

```
Hint: Your generator might be starting with the wrong (and non-prime number).

Also, _music.py_ fails to run. Make sure it runs correctly. Should be an easy fix.

Thanks,

Luis

ldconejo:
Great! Also, _music.py_ still fails to run. Look at your code on lines 16 & 17, should be a very easy fix.

-Luis

PR #79, opened 2018-07-08
syslog simple.py

alexLaws:
Logging assignment. Hard to tell if it worked 100% because the output is screwy for windows but I think it did.

### DIFF
--- a/students/alexLaws/lesson05/simple.py
+++ b/students/alexLaws/lesson05/simple.py
@@ -1,0 +1,47 @@
+#!/usr/bin/env python3
+
+import logging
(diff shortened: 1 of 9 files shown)


alexLaws:
This includes the unit testing calculator assignment now.

ldconejo:
Hi Alex,

The output is not very pretty on Windows, it's a side effect of having to use the DatagramHandler. Other than that, it sends remote error log entries as expected. Please remember to include all collaterals (such as syslogserver.py) as part of your commit. In this case, I used a copy of syslogserver that I already had.

PR #80, opened 2018-07-09
Lesson8 - Assignment1

ghassan24601:


ldconejo:
Good job. A couple of changes I would make for this type of assignments would be to make sure the directory structure is preserved in the commit (i.e., there was no ../logs directory, so someone trying to run the code would have to figure out that's missing) and maybe include _config.ini_, populated with stubs instead of your passwords, so that someone testing the code only needs to update the passwords themselves.

ghassan24601:
Thank you for pointing that out. I actually have made the same mistake at work :)

PR #82, opened 2018-07-11
Submitting Lesson 04

wesmoy17:
See mailroom_oo.py


ldconejo:
Hi Wesley,

For this assignment, one of the goals is that the data saved in JSON format can be both saved and loaded. I see you implemented a load method, but it's not being used. Could you add that functionality?

Also, in terms of real-world applications, it would probably make more sense to have your save method place all donors into a single JSON file, instead of creating individual ones for each donor.

Thanks,

Luis

PR #89, opened 2018-07-15
lesson07 part 2 - adding mailroom files (2)

kmsnyde:
Run mailroom_op.py.  Database is pre-loaded.  In menu option 1, you can type a new name to add and new donor and donation.  In option 4, you can delete a donor.

### DIFF
--- a/students/kmsnyde/lesson07/mailroom_op.py
+++ b/students/kmsnyde/lesson07/mailroom_op.py
@@ -1,0 +1,148 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Sun Jul 15 12:20:46 2018
+
+@author: HP-Home
+"""
+import logging
+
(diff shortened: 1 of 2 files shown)


ldconejo:
Hi Karl,

For this assignment, you need to enable capabilities to add / update / delete data from the database. The first one is completed by running _mailroom_sql.py_, which it turn runs _populate_db()_. Now, you need to implement the capabilities to update that data (could be either changing the name of a donor or updating/adding to the amount donated) and delete (such as deleting a donor). The way _mailroom_op.py_ works is perfect for this, you would just need to make the data persistent (that is, having donor deletions or donation updates uploaded to the database). For now, I will close this pull request, you can submit a new one (just on Github, the one on the Course's website will stay open) once those changes are done. 

Thanks,

Luis

PR #100, opened 2018-07-19
Lesson07 Mailroom Update

kmsnyde:
Removed the code that made it seem as though the database data was not persisting. It is clear the data persists in donor.db at this time.


ldconejo:
Hi Karl,

Thanks, it looks good now. One thing to consider for a real-world application (although not an issue for the purposes of the assignment) is that when you delete a donor, you're deleting that donor from the _donor_ table but not that donor's donations from the _donation_ table. Thank you for your patience in submitting multiple pul requests for this assignment.

-Luis

PR #103, opened 2018-07-20
Context managers request.

hortw1991:


hortw1991:
Lesson003 - context managers and recursive factorials

PR #109, opened 2018-07-24
completed water regulation assignment

rmart300:
few flake8 warnings remaining


ldconejo:
Hi Stephen,

Everything looks good, however pylint has an error about using an invalid-name, among other things, Flake8 also complains about the length of some code lines . Try running both, look at the warnings and see if you can make some code improvements based on them. 

Thanks,

Luis

PR #114, opened 2018-07-27
lesson04 [very beta]

johnpharmd:


ldconejo:
Hi John,

Since you're generating _donorgroup_ from _self.donors_ every time, it's not really necessary to have it as a property, you could get away with removing the _property_ and _setter_ decorators and adding a print statement to _donorgroup,_ which is more or less what you're doing with _get_report_:

```
    def donorgroup(self):
        """getter for self.donors; returns list of donors in group"""
        print(sorted([donor for donor in self.donors]))

    def donorgroup_new_donor(self, new_donor):
        """enables addition of new donor last name to donorgroup list"""
        self.donorgroup.extend(new_donor.donor.keys())
```

Now, if you do want it as a property, one possible way would be doing this:

```
class DonorGroup:
    """creates donor group dictionary objs for multiple donor dictionaries"""
    def __init__(self, donors):
        self.donors = donors
        self.donorgroup = sorted([donor for donor in self.donors])

    @property
    def donorgroup(self):
        return self.__donorgroup

    @donorgroup.setter
    def donorgroup(self,donorgroup):
        self.__donorgroup = donorgroup

    @donorgroup.setter
    def donorgroup_new_donor(self, new_donor):
        """enables addition of new donor last name to donorgroup list"""
        self.__donorgroup.extend(new_donor.donor.keys())
```

Note that in this case, I'm using a couple of setters, one for simply assigning a list to it and another for simply extending the list of donors with the name of a new donor. Hope this helps!

-Luis

PR #116, opened 2018-07-28
Dustin_L - Lesson07 Mailroom

rdlaner:
Completed Lesson07 Mailroom assignment.


ldconejo:
Hi Dustin,

The code looks good and the functionality to add data to the database is clearly there. Could you also add the capability to update and delete data? For updating, it could be the amount of a donation for a given donor. For deleting, it could be someone in the list of donors (see if you can also make it so that donations from that donor are deleted as well).

Thanks,

Luis

PR #126, opened 2018-07-31
Lesson02 - Generators and Closures

DeFeNdog:


ldconejo:
Hi Carlos,

Your generator looks. For the enclose function, the way it is currently coded, it will only return the first track that it finds that matches the energy level condition. Please fix the code so that it returns all high energy tracks. You can achieve this by having the enclosed function behave as a generator (easiest change) or you could modify it to return a list.

Thanks,

Luis

PR #134, opened 2018-08-02
Lesson09|Threaded Scraper

nDruP:
I replicated the code from the video to get the appropriate news sources and articles.
The code for threading/queues was also very similar to the GIL video.
I would have liked to get some practice using the asyncio module, but I'm sure it'll happen in the future.

### DIFF
--- a/students/nDruP/lesson09/threaded_scraper.py
+++ b/students/nDruP/lesson09/threaded_scraper.py
@@ -1,0 +1,86 @@
+"""
+Scrape the newsAPI using threads instead of the async approach
+Create a pool of threads and send a request to the pool for processing?
+https://newsapi.org
+
+EX: NewsAPI request Top business headlines in the US right now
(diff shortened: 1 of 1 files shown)


ldconejo:
Hi Andrew,

I'll capture your input on getting more practice on the asyncio module, I think it's a good idea to eventually include a second assignment for this lesson.

Regards,

Luis

ldconejo:
Hi Andrew,

Looks good. There's an error near the end, when you do:

`if (article['title'].lower()).contains(SCRAPE_WORD):`

_str_ doesn't have a _contains_ method, so you'll need to use something different to see if your scrape word is in any of the received headlines. I did run with a fix for that line, funnily enough, there were no matches for the word "justice". "money" did get two or three hits :-)

Please send a new pull request once that is fixed.

-Luis

nDruP:
I changed it to 
**SCRAPE_WORD in article['title'].lower()**

PR #141, opened 2018-08-04
complete with metaprogramming of mailroom

etwum:
added save and load methods


ldconejo:
Hi Emmanuel,

Project looks good, when the save function is used, only the last donor entered is saved. Could you fix it so that all donors are saved to JSON?

Thanks,

Luis

etwum:
Here's the new pull request

https://github.com/UWPCE-PythonCert-ClassRepos/SP_Online_Course2_2018/pull/142

PR #143, opened 2018-08-05
lesson09 newsapi

kmsnyde:
just fyi, very frustrating trying to work the code against "newsapi" as my usage limit is reached quickly and I cannot interact with the site to further verify code until many hours later.

### DIFF
--- a/students/kmsnyde/lesson09/get_news_threaded.py
+++ b/students/kmsnyde/lesson09/get_news_threaded.py
@@ -1,0 +1,65 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Fri Jul 27 16:28:49 2018
+
+@author: HP-Home
+"""
+import requests
+
+from threading import Thread
+
+WORD = 'trump'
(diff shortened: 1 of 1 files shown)


ldconejo:
Hi Karl,

Yes, working with threads can be frustrating at times. I would suggest you limit the number of news sources to two or three, so that you don't bombard the API with requests, running outside of usage limits for the free tier.

Aside from that, I'd suggest the following changes:

- Don't do a threaded call to get the sources, it's not necessary, as _get_sources()_ returns a single list of news sources. One problem in your code right now is that you're not doing a .join after launching _get_sources()_ as a thread, so the rest of the code runs without waiting for _get_sources()_ to put anything into _sources_.
- Once you have a list of news sources, THEN you launch one thread of _get_articles()_ per source is _sources_ (which your code is already doing, although there is an error there). To avoid running into usage limits, before launching one thread per source, do something like _sources = sources[:3]_, which will trim the list of sources to only the first three elements. You can try the full list once you know your code is working.

Happy coding!

Luis

PR #148, opened 2018-08-07
mailroom program operational with sqlite backend

rmart300:


ldconejo:
Hi Ross,

Implementation of the database looks good, I was able to add new donors and their donations. As per the assignment requirements, please add the capability to update data (could be either updating a donation amount or the name of a donor) and delete data (either a donor and his/her donations or an specific donation).

Thanks,

Luis

ldconejo:
Hi Ross,

This is great! Not only did you add update and delete capabilities, the implementation you made on the user interface is is both simple and easy to use.

Regards,

Luis

PR #151, opened 2018-08-08
Update to lesson09

kmsnyde:
First, thanks for the tip to use sources[:3] prior to running the thread as it did not even occur to me when I was having the issue of using all of my 'free" requests from newsapi.

I found my error in the thread, which you indicated I had one but thanks for not telling my as i figured that out, as well (args need a tuple).

I took out the Thread to get_sources; thanks as I realize that process prob doesn't need a thread.

program seemed to work now without errors.  Thanks for the guidance.


ldconejo:
Hi Karl,

I'm glad you found the feedback useful. As you said, the program now without any issues. Happy coding on the final assignment.

-Luis

PR #153, opened 2018-08-09
lesson_07 submission

slickw1lly:


ldconejo:
Hi Kim,

The department implementation looks good. For the second assignment, could you add capabilities to update/delete information? One suggestion is to allow deleting a donor (and his/her associated donations) and also allowing to update a donor's name.

Thanks,

Luis

PR #163, opened 2018-08-12
lesson04 json_save_meta

johnpharmd:


ldconejo:
Hi John,

Some clues as to what you need to do next:

1. Import json_save_meta.py
2. Make your _DonorGroup_ class inherit from _JsonSaveable_ (which is within _json_save_meta_).
3. Create a save method within _DonorGroup_ which will be in charge of saving the data to a json file. You'll also need a load method to get the data from the json file, instead of having your donor dictionary hardcoded. 

Let me know if that helps.

Regards,

Luis

PR #167, opened 2018-08-13
lesson04 mailroom

johnpharmd:


ldconejo:
Hi John,

Is this a commit for review of the assignment or are you still debugging it?

Thanks,

Luis

johnpharmd:
Hi Luis, it still has the attributes issue we were discussing.

John

ldconejo:
Hi John,

Try the following changes to your _DonorGroup_ class:

1. Add a class variable indicating a saveable attribute, for example:

`donors = js.Dict()`

Note that _Dict()_ is defined in _saveables.py_.

2. Remember to import the _json_ module.
3. You don't need that _infile.read()_ for handling a json file, use instead _json.load()_
4. Remember that if you want to call _load_data()_ you need to do it for the instance, so it should be _self.load_data()_

You'll probably run into a few more challenges, keep me posted on your progress!

Thanks,

Luis

ldconejo:
Here's a code example from another student that you might find useful:

`class DonorDatabase(js.JsonSaveable):
    """A database of Donors"""

    _donors = DonorDict()
    min_col_width = js.Int()
    def_pad = js.Int()
    col_sep = js.String()
    cols = js.List()
    thank_you_fmt = js.String()

    def __init__(self, *donors):
        if not donors:
            donors = []

        self._donors = DonorDict(*donors)
        self.min_col_width = 12
        self.def_pad = 7
        self.col_sep = ' | '
        self.cols = ['Donor Name', 'Total Given', 'Num Gifts', 'Average Gift']
        self.thank_you_fmt = ('\nDear {:s},\n'
                              'Thank you for your generous donation of ${:,.2f}.\n'
                              '\t\tSincerely,\n'
                              '\t\t  -Your conscience')`

PR #168, opened 2018-08-14
Lesson 07 - Department

etwum:
added a department class
added data to the database
created a print file

### DIFF
--- a/students/etwum/lesson07/create_personjobdept.py
+++ b/students/etwum/lesson07/create_personjobdept.py
@@ -1,0 +1,133 @@
+"""
+    Simple database example with Peewee ORM, sqlite and Python
+    Here we define the schema
+    Use logging for messages so they can be turned off
+"""
+
+import logging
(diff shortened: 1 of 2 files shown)


ldconejo:
Hi Emmanuel,

Were you able to run this code to completion? In my case, it gets stuck when trying to the create the _Department_ table. There's an error in the way you're entering the constraints for _department_number_, check that and also see if you can make the constraint simpler (for example, by using _BETWEEN "A" AND "Z"_).

Thanks,

Luis

etwum:
I wasn't getting an error. It would just run through the log statements
then stall out. Probably the reason the database showed no data. I wasn't
sure if it was because I didn't have SQLite installed properly. Anyways it
took awhile but I figured it out, it was because I didn't have brackets
around Check.

New pull request:
https://github.com/UWPCE-PythonCert-ClassRepos/SP_Online_Course2_2018/pull/175

PR #184, opened 2018-08-20
Submitting Mailroom Assignment

wesmoy17:
Submitting Mailroom Assignment for review. Also, I attached the wrong file when submitting the last assignment. The website says I cannot resubmit a link/file anymore. Is there anything that can be done to undo this?


ldconejo:
Hi Wesley,

Don't worry about posting the wrong link. I usually follow pull request notifications, which go separately from what's posted on the course website.

On this assignment, could you add capabilities to update and delete data? One option would be to allow changing the name of a donor or changing the amount of an existing donation. The same goes for the delete capability, it could be deleting an existing donor or a specific donation.

Thanks,

Luis

PR #186, opened 2018-08-20
[DennisLee][L4][Mailroom Metaprogramming] Submission.

DennisLee-DennisLee:
It appeared that the assignment description only required that the donor/donordb class be modified to save and load JSON files properly, so I added that functionality without updating the donor UI or any test files. Let me know if I need to fix those as well.


ldconejo:
Hi Dennis,

It's ok. Having an updated UI does help a reviewer in verifying your code to be working as expected, but the demo that executes with _mailroom.py_ is more than enough to see that the code you added indeed saves and loads data in JSON format.

Regards,

Luis

PR #187, opened 2018-08-21
Lesson08/mailroom

JerHsi:


ldconejo:
Hi Jerry,

The assignment looks good. For the Redis implementation, could you enable the capability to lookup for data? As suggested, you can implement lookup by email, but any other fields would also be fine.

Thanks,

Luis

PR #208, opened 2018-08-29
Testing Assignments

hortw1991:
I was unable to get pylint to ignore duplicate code that I didn't want to make a global variable or share between modules.


ldconejo:
Hi Tom,

The calculator assignment works fine. For the water regulation assignment, you're getting to 9.78 on Pylint and the duplicate code is on the test, so that's ok. However, you integration test isn't really running any tests:

```
----------------------------------------------------------------------
Ran 0 tests in 0.000s

OK
```

Can you figure out why? It's an easy fix.

Thanks,

Luis

PR #219, opened 2018-09-02
[DennisLee][L5][P2: Logging] Submitting.

DennisLee-DennisLee:
I think this is ready to go, although for some reason the server only responds when I'm in debugging mode. Also, I've seen where the expected output doesn't appear until about a minute later, after I rerun the test program later in the same debugging session. I wasn't able to unclutter the output, either.


ldconejo:
Hi Dennis,

I see log activity right away on my setup (I'm running on Mac OS X), both the remote log as well as the local one. The convoluted output is a "feature" of DatagramHandler.

-Luis

PR #223, opened 2018-09-05
I'm striking out on this. Any assistance you can provide would be appreciated

MikeS-1:


ldconejo:
Hi Mike,

This was an interesting one. Two things to do:

1. Install a database inspector and look at the _personjobdept.db_ your code is creating. You'll confirm that your code is already doing what it is supposed to do. Kudos on that!

2. In the database inspector, look at the data entered into the _Job_ table, in particular the column names. Verify those column names match your model in _personjobdept_model.py_. Based on what you see there, it's possible you need to modify your code where you're querying the job data you just entered.

Once you've done what's listed above (not before), take a look at this discussion, it more or less explains what happened:

https://github.com/coleifer/peewee/issues/44 

Hope that helps,

Luis

PR #226, opened 2018-09-05
pandas doesn't work... incomplete codes

HiroTestStudent:
Hi Luis,

Pandas causes the error. Error message is below. 

Hiroyukis-MacBook-Pro-2:Lesson1 hiroyukitakechi$ python comprehensions.py
/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/importlib/_bootstrap.py:205: RuntimeWarning: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88
  return f(*args, **kwds)
/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/importlib/_bootstrap.py:205: RuntimeWarning: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88
  return f(*args, **kwds)

Additionally, my sorted() function has an issue too. This is my issue....

Thanks,
Hiro

### DIFF
--- a/students/HiroyukiTakechi/Lesson1/comprehensions.py
+++ b/students/HiroyukiTakechi/Lesson1/comprehensions.py
@@ -1,0 +1,17 @@
+'''
+Lesson 1 Assignment #1
+'''
+
+import numpy as np
+import pandas as pd #this code gives me an error message
+music = pd.read_csv("featuresdf.csv")
+#music.head()
+#music.describe()
+
+#[x for x in music.danceability if x > 0.8]
+
+results = [(a,d,l) for a,d,l in zip(music.artists, music.danceability, music.loudness) if d > 0.8 and l < -0.5]
+
+results2 = sorted(results, key=results.danceability) #this code is an issue, too.
+results.head()
+
--- a/students/HiroyukiTakechi/Lesson1/generator_solution_oo/generator_solution.py
+++ b/students/HiroyukiTakechi/Lesson1/generator_solution_oo/generator_solution.py
@@ -1,0 +1,32 @@
+'''
+Lesson 1 Generator Draft 1
+'''
+
+
+def intsum():
+    x = 0
+    while True:
+        yield x
+        x += x+1 #this is not right.... I need to think a little more.
+
+
+def doubler():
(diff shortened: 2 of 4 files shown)


ldconejo:
Hi Hiro,

As you correctly assumed, I don't see the error with Pandas when trying to run your code. It seems you're seeing a similar issue as listed here:

https://stackoverflow.com/questions/40845304/runtimewarning-numpy-dtype-size-changed-may-indicate-binary-incompatibility 

Apparently, it's collateral damage for using a version of scipy that was compiled with an older version of numpy. Try reinstalling scipy by following these instructions:

pip uninstall -y scipy scikit-learn
pip install --no-binary scipy scikit-learn

As for the other issue, take a look at the explanation here:

https://www.pythoncentral.io/how-to-sort-a-list-tuple-or-object-with-sorted-in-python/ 

Look for _Sorting a list of tuples_, that's what you need to do in this case: Create a function that will extract the danceability value from each tuple and feed that function as key to the sorted function. I can show you how to do that directly when you create your _results_ variable. We can talk more about it during student hours.

-Luis

HiroTestStudent:
Hi Luis,

I just tried to uninstall and reinstall scikit-learn by following your instruction but still it shows the error message below. I am wondering if I need to reinstall numpy too. I will talk to you during office hours at 8:30pm. Thanks, Hiro

Hiroyukis-MacBook-Pro-2:Lesson1 hiroyukitakechi$ python comprehensions.py
/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/importlib/_bootstrap.py:205: RuntimeWarning: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88
  return f(*args, **kwds)
/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/importlib/_bootstrap.py:205: RuntimeWarning: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88
  return f(*args, **kwds)

ldconejo:
I found something about it: https://github.com/numpy/numpy/issues/11788 

What they say is to ignore the warning for now. Let's take a look during office hours.

-Luis

PR #227, opened 2018-09-06
adding lesson06 calculator files and tests

natrod22:


ldconejo:
Hi Natalie,

A few changes you need to make:

- Add a __init__.py inside of the _calculator_ folder to turn it into a package.
- Delete the copies of _adder.py_, _calculator.py_, etc. that exist outside of the _calculator_ folder.
- In both _unit-test.py_ and _integration-test.py_, change your imports to reflect the fact that _calculator_ is now a package. For example: 

`from exceptions import InsufficientOperands --> from calculator.exceptions import InsufficientOperands`

- Make sure all your tests are passing.

Let me know if you have any questions,

Luis

PR #231, opened 2018-09-06
Generator_Solution Prime() was fixed

HiroTestStudent:
Hi Luis,

Thank you for help last night. I fixed the prime() in the generator_solution.py file.

Thanks,
Hiro


ldconejo:
Hi Hiro,

Great job! For _comprehensions.py_ an easier way to only get (and print) the first five results is this one:

```
for result in results[:5]:
    print(result)
```

_results[:5]_ effectively returns the first five elements, so all you need is to print them out.

Good luck with your next assignment!

Luis

PR #246, opened 2018-09-12
HW 1 & 2

nebiatab:
Rest of HW1, and HW 2


ldconejo:
Hi Nebiat, 

Assignment 2 looks good. For assignment 1, please include _test_generator.py_ (modified as required to test your code). Also, to keep things in order, please create separate folders for each assignment ("lesson01", "lesson02", etc.). Some of the following assignments have more elaborate directory structures, son it's best to keep each lesson/assignment in a different folder.

Thanks,

Luis

nebiatab:
Thanks for the feed back. I've added in the test file and organized each hw to have its own folder.

PR #247, opened 2018-09-13
Part 1 Commit.  All work locally. 

hortw1991:
I had to edit some file paths which randomly stopped working.  Working on fixing them back to what they were, since I'm guessing it will mess things up on your end.


ldconejo:
Hi Tom,

learnnosql.py should be able to call all three databases (MongoDB, Redis and Neo4J). Right now, your code only running a few operations locally. Let me know if you have any questions.

Thanks,

Luis

PR #249, opened 2018-09-14
Herela: Lesson 7!

harvey-herela:


ldconejo:
Hi Harvey,

Great job on the first part! For the second part, please add the functionality to update and delete data from the DB. One option is to allow updating a donor's name. By the same measure, you can allow deleting a donor and all of his/her donations.

Regards,

Luis

PR #268, opened 2018-09-23
lesson08 - PR to get feedback on hmset, hmget attempts in mailroom_redis

johnpharmd:
Hi Luis,
I'm sure I'm trying something that legions of previous learners have learned not to do the hard way,
but am trying to create lists of r.hmset and r.hmget in string format, then run the r.hmset and r.hmget commands by stripping the quotation marks.

Doesn't crash my program, but also doesn't return desired output. Is this hopeless, ie, should I just type out each r.hmset and r.hmget command? Or, if it can be done, what should I do differently?

Thanks, as always, for your help. 
John


ldconejo:
Hi John,

It's definitely a harder approach. What you're trying to do is to execute code that exists on a string. It should be possible to do it in a way in your _hmget_ and _hmset_ statements are not part of the string and are instead composed at execution.

That said, it is possible to execute Python code from a string. You can look at this discussion, it has some code that would be usable:

https://stackoverflow.com/questions/701802/how-do-i-execute-a-string-containing-python-code-in-python 

Let me know if that helps. I will close this pull request, but feel free to reopen it as required.

Regards,

Luis

PR #271, opened 2018-09-24
lesson08: Fixup adding donations

mayc4t:


ldconejo:
Hi Lily,

Good job on adding additional capabilities to all three DB implementations. The way to stop adding new donations (typing "quit" as last name) was a bit out there, but still usable :-)

Regards,

Luis

PR #273, opened 2018-09-25
Help with L7 Department - query not showing external table fields.

DennisLee-DennisLee:
The job stint query in **DennisLee_db_manipulation.py** executes as expected, and even sorts correctly based on a Department table field despite selecting mainly from the Job table. However, if you cycle through the query results, you'll get an error if you comment out the lines that reference the Department fields (see lines 227 and 228). Can you help figure out what the code bug is there?

To run the code, just invoke **DennisLee_db_manipulation.py** or call ManipulateDb() from that file.

### DIFF
--- a/students/DennisLee/lesson07/department/DennisLee_db_manipulation.py
+++ b/students/DennisLee/lesson07/department/DennisLee_db_manipulation.py
@@ -8,12 +8,11 @@ import logging
 import peewee as pw
 import DennisLee_model as mdl
 
-class manipulate_db:
+class ManipulateDb:
     """
     This class creates the database and performs database transactions.
     """
     def __init__(self):
-
         self.db_name = 'DennisLee.db'
 
         self.set_up_logging()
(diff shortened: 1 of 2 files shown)


ldconejo:
Hi Dennis,

I'm still at a loss on this one. Your schema does work correctly, as I'm able to see it on a DB inspector and run the following query on it (almost the same as what you're trying to do with Peewee, except for the duration part):

`SELECT Job.person_employed_id,Job.job_name_id,department.department_name,department.manager
FROM Job 
JOIN deptjobs ON Job.job_name_id=deptjobs.job_name
LEFT OUTER JOIN department ON deptjobs.department_number_id=department.department_number`

Which results in this:

image

I made a few changes to your code, including separating the join statement for each table, but the joined columns still do not show up:

`        data_query = (mdl.Job.select(
            mdl.Job.person_employed_id,
            mdl.DeptJobs.job_name.alias('DeptJob_name')
        )
        .join(mdl.DeptJobs)
        .join(mdl.Department, pw.JOIN.LEFT_OUTER, on=(mdl.DeptJobs.department_number == mdl.Department.department_number)))
        for gig in data_query:
            self.logger.info(vars(gig))
            self.logger.info(
                f'Person {gig.person_employed} '
                f'DeptJobs name {gig.DeptJob_name} '
            )`

For assignment purposes, you could make the _department_number_ a foreign key in the _job_ table, or even modify the _department_ table, so that the _department_name_ becomes the new primary key and can be imported as a foreign key into the _job_ table, so that you don't need to do a join. I'll take another look at your code later to see if there's something else causing the join operation not to work as expected.

Thanks,

Luis

PR #291, opened 2018-10-01
Lesson 09.

hortw1991:
Seemed to work, then started getting error messages related to the number of requests available for free accounts making testing a little spotty.


ldconejo:
Hi Tom,

Your implementation is correct. When debugging projects like this one, a good trick is to limit the amount of sources being queried, as the free modalities of these apps often have limitations. You could try something like this:

`for source in sources[:5]:
    jobs_queue.put(source)`

Which limits the number of sources polled to just 5.

Good job on this assignment,

Luis

PR #299, opened 2018-10-05
Committing logging section of assignment 5. Git will not allow me to …

letoui:
…add .log files to git. Posting in assignment submission.


ldconejo:
Hi Ian,

For the debug log, it's ok that included it in your submission. However, in the course's website, the debug log part is a separate submission, make sure you submit it there, so that I can grade it. 

For the logging assignment, the code will currently crash because some missing elements when you instantiate your _SysLogHandler_, could you fix that?

Thanks,

Luis

PR #310, opened 2018-10-10
Assignment 2 update

Swt16:
Assignment 2 update.


ldconejo:
Hi Shin,

For the generator, the way you're coding it will turn the list of tracks into a list. Two clues:

1. Using [] in comprehension will turn the result into a list.
2. You don't need to sort the results for this one.
3. Once _fav_ becomes a generator, you will be able to do _next(fav)_ to verify it.

Happy coding,

Luis

PR #311, opened 2018-10-10
Second half of assignment 6

letoui:


ldconejo:
Hi Ian,

For this one, the second assignment is working correctly, you need to do a few changes on the first part:

1. Unit test and integration test have to be callable from outside of your calculator project (you're using the name _advanced_testing_ which is fine).
2. Make sure you don't have any significant warnings from Pylint or Flake8.

Thanks,

Luis

letoui:
Hey Luis,

I’m not sure I entirely understand what you mean for number 1. Am I missing something in the test files themselves or am I just not calling the test files correctly?

Thank you,
Ian Letourneau

ldconejo:
Hi Ian, 

Sure. The request on #1 refers to this:

_Note that all of the command examples below should be run from inside the project root which contains the unit-test.py and integration-test.py files_

In the case of your submission, your unit-test.py and integration-test.py files should be one level up from where your calculator files are. One reason for this is that, when checking coverage for this part of the assignment, we will run this command

`python -m coverage run --source=advanced_testing -m unittest unit-test.py`

When doing a coverage run with _--source_ all files in the directory will be included in the coverage analysis, so you generally want the unit and integration tests out of that.

Let me know if that makes sense.

Thanks,

Luis

PR #312, opened 2018-10-11
Assignment #2 from cindy waldron

cindywaldron:


ldconejo:
Hi Cindy,

The idea of a closure is that you return a function which has a preloaded parameter. In this case, that preloaded parameter is the energy level, which you're setting at 0.8. Right now, you code is returning the output of _energy_filter_ when using _music_ as parameter. You will need to change that so that it returns _energy_filter_ only. The last two lines of your code should look similar to this:

`energy_closure = energy_track_gen(0.8)
print(energy_closure(music))`

Let me know if that makes sense.

Thanks,

Luis

PR #313, opened 2018-10-11
take two of Lesson 5 Logging assignment

HiroTestStudent:
Hi Luis,

Thank you for help during office hours. Now I think I have a better understanding what the codes do. I refactored the codes and re-ran on Terminal. It worked. FileHandler created the file too. Since I still can't make sudo works on my Terminal, I couldn't test the remote but I assume it will work as you showed me during office hours.

Thanks,
Hiro


ldconejo:
Hi Hiro,

You made it work, that's great. Remember that you need to submit on the course's website so that I can grade your submission.

Thanks,

Luis

PR #314, opened 2018-10-11
Lesson 08 - Problems with MongoDB in NOSQL and the mailroom

DennisLee-DennisLee:
Mainly concerned with NOSQL at the moment - whatever the MongoDB fix is will probably fix the mailroom problem as well.

I can log in to the MongoDB database, but when I try to insert data, the idna codec problem surfaces.


ldconejo:
Hi Dennis,

A couple of things that you can start trying:

1. Within _mailroom_oo.py_, add exception handling to see if your query, when checking a person's hometown is returning valid data. That will take of the error you're getting at this point. You will then need to figure out the next error (you're trying to call a _transaction()_ method that doesn't exist in the database object:

`        query = self.persons.find_one({'person_name': clean_name})
        try:
            donor_town = query[0]['lives_in_town']
        except:
            donor_town = 'no_town'
        if clean_town and query and donor_town != clean_town:
            self.logger.info(
                f"Person {clean_name} is in table but with another hometown "
                "update the record to change to the new hometown.")`

2. Try to use context managers when you're accessing the database, your code is currently running very slowly. Try things like this:

`    with login_database.login_mongodb_cloud() as client:
        db = client['mr']
        persons = db['Person']`

That will take care of closing open instances of the database when you finish using them.

Let me know if this helps.

Regards,

Luis

DennisLee-DennisLee:
I’m not able to access my course computer at the moment, but I’m fairly certain that #1 won’t help solve the problem. The idna issue occurs the first time I try to write to the DB (to fill it with data), well before I would try to read from the DB.

For #2, I think I’m already doing that in the NoSQL code. I’ll consider it for the mailroom, but it would probably require a lot of code restructuring.

My NoSQL code is all but finished, and there’s a reasonable chance that everything will run fine once I get past the idna problem. The mailroom MongoDB code is in rougher shape, and even after I fix the idna problem, I expect there will be at least a few other issues to correct. I was hoping you’d look at the NoSQL stuff without getting distracted by the mailroom code, but unfortunately I had already committed and pushed the mailroom bits to the server before I made the pull request.

ldconejo:
Hi Dennis,

I think I've found the issue with your code, but I'm curious as to why you're doing it this way: You're transforming the connection string for MongoDB using _urllib.parse.quote_plus_. Try changing the connection string to feed it directly into _pymongo.MongoClient_, something like this:

`client = pymongo.MongoClient(f'mongodb://...`

That will get you out of the IDNA error. You can also print the connection string after being processed by _urllib.parse.quote_plus_ and see why that is causing so much trouble.

Let me know if that works,

Luis

DennisLee-DennisLee:
I had tried the login as the Git repo example illustrated, but it wouldn’t let me connect. The exception message said to use urllib.parse.quote_plus, so I added it to the code. After that, the connection login worked, but that’s when I started running into the IDNA problem downstream (I’m able to create a database collection, but get the error when I try to fill the collection with data).

My MongoDB username is an email address, so maybe the cloud server didn’t like the at-symbol and period being in the connection string. Should I create a new MongoDB account without an email address? I’m not sure how to do that – the email seemed to be required. Also, the rest of the connection string has an @ symbol, so the username might not actually be problematic.

ldconejo:
Hi Dennis,

I see. The user name you use in the connection string is not the email address you used to create your account, but the one you created for the user you made for the cluster. Do the following:

1. Go to https://cloud.mongodb.com/user#/atlas/login and login with your email and password.

2. Once you are logged in, you will be presented with your list of active clusters (I assume only one):

image

3. Click on the _Security_ tab, that will show existing users, their access method and level of access. _This_ is the user name you need for your connection string.

image

4. Keep in mind that each user has their own password, different from the one you use when accessing the MongoDB website. If you don't remember that one, edit the user and generate a new password.

I will close this pull request for now, please reopen it if required. Also, if you're still blocked, maybe we can setup a quick office hours session to debug your code together.

Regards,

Luis

DennisLee-DennisLee:
Thanks, this sounds like it’ll fix the issue. I’ll try it after I finish work tonight.

PR #331, opened 2018-10-17
Lesson09 - Assignment

DeFeNdog:


ldconejo:
Hi Carlos,

The way you're running your threads right now, you're running _count_word()_ before the threads have finished, so you're essentially checking an empty _TITLES_ list, with zero matches every time. You need to make sure your threads have finished first. Try the following:

In _threading_articles()_ created an empty _threads_ list, within the for loop you're using to create your threads, add a _threads.append(thread)_ at the end of each loop, this will save a pointer to each thread within the _threads_ list. 

Once you're done with the for loop, wait for the threads to finish by doing something like this (this also within the same _threading articles_ function):

`for t in threads:
    t.join()`

This will ensure that the _threading_articles_ function will not return until all threads are finished.

Once you do that, you'll discover other errors in _count_word_, I'll let you figure those out ;-)

One final recommendation: Try to have a _main_ function where your execution occurs and then finish your code with this:

```
if __name__ == '__main__':
    main()
```

In this way, you keep execution contained within a procedure, instead of execution and function definitions being intermingled. Let me know if this makes sense.

Regards,

Luis

DeFeNdog:
Hi Luis,

I'll attempt the suggestions you made. Are you seeing anything in my code
that would suggest I'm making an abnormal number of requests? I hit the
usage limit pretty quickly on newsapi.org. That's the main reason I
couldn't finish proper testing with my code.

Thank you,

Carlos

PR #333, opened 2018-10-17
Lesson 6 Advanced Testing assignment: just adding two classes...

HiroTestStudent:
Hi Luis,

I am working on the Lesson 6 Advanced Testing assignment but all I did so far is to add two test classes being instructed. But unittest doesn't work... 

Thanks,
Hiro


ldconejo:
Closing as code has already been discussed during office hours.

PR #336, opened 2018-10-18
Final version of Lesson 5 Advanced Testing assignment

HiroTestStudent:
Hi Luis,

I fixed codes of the lesson 6 advanced testing assignment.

Thanks,
Hiro


HiroTestStudent:
Hi Luis,

I tried but I got even worse score so I gave up...

I will come back to this assignment later. I will take a look at water regulator assignment first. In case I should fall behind, I will move onto lesson 7. I will take one week off in Japan from Nov 3 to 10 so I prefer not to fall behind by then.

Thanks,
Hiro

PR #347, opened 2018-10-23
Assignment 4 files for debugging

Swt16:
Assignment 4 files for debugging


ldconejo:
Hi Shin,

You were actually very close. In _mp_mailroom.py_, line 78 should not be treated as a list:

`@js.json_save
class DonorCollection:

    donor_dict = js.List()  --> This one should be instead: donor_dict = js.Dict()

    def __init__(self, donors):
        self.donor_dict = donors`

Once you have made that change, trying saving to JSON again.

Regards,

Luis

PR #352, opened 2018-10-25
Assignment 4 Updates

Swt16:
I can't see to get my load from json method to work properly. It doesn't update anything. I tried changing the syntax around, and it says that either the dictionary doesn't have any items, or the object can't call from_json_dict. Any idea what I need to change in my code?


Swt16:
Hello Luis,

I think you were looking at my old version, either way, you pointed me in the right direction, got it to work.

This was my unfinished code that I was asking about:
```
        """Loads donor information from a json file"""
        with open('Data1.json','r') as data_load:
            json_data = json.load(data_load)
        donor_dict = self.from_json_dict(json_data)
        print("Donors have been loaded.")
```

And this is my solution:
```
        with open('Data1.json','r') as data_load:
            json_data = json.load(data_load)
        self.donor_dict = self.from_json_dict(json_data).donor_dict
        print("Donors have been loaded.")
```

Thanks,
Shin

ldconejo:
Hi Shin,

Interesting, do you mean I was looking at a previous pull request or at the wrong code within the correct pull request? In any case, I'm glad your code is working, I just reviewed your latest pull request. Good luck on the next assignment!

Regards,

Luis

Swt16:
Yeah, you were looking at the previous pull request/wrong code. I had some minor trouble with Git cause it said there was a pull request pending, so I had to manually update the code on this website.

Thanks,
Shin

PR #368, opened 2018-11-01
lesson 5 assignment

cosmonautz:


ldconejo:
Hi Kevin,

Thanks to your implementation, I discovered that I didn't have a value for _localhost_ setup in my /etc/hosts. Most other students have entered the localhost default address (127.0.0.1) in their implementations, so at first I thought this was an issue with your code (using _localhost_ is completely valid, by the way). Interesting one :-)

Thanks,

Luis

cosmonautz:
Sorry about that, I was doing the assignment on Windows but somehow skipped
over the part about using DatagramHandler at first so was beating my head
against the wall using the SysLogHandler and trying everything to get it to
work.  First change I made was to use 127.0.0.1 but after that didn't work
i tried localhost.  When I eventually realized my mistake and switched to
DatagramHandler (and it worked) i just turned it in.  Had i just read the
assignment closer i probably never would've tried using localhost.

PR #369, opened 2018-11-01
Session01

toddlikesbeer:
Hey new teacher folk,

I had some issues setting up Git, so i was unable to make incremental commits this session. I will try and be better about that in future lessons. 

I was also a but confused on what I was supposed to do for Iterators & Iteratables assignment; it wasn't exactly clear. So if there is something else I need to do please let me know and I will knock it out. 

REALLY looking forward to the class. Do ya'll have a Slack instance like the Python 1 class or is email the best way to reach you?

### DIFF
--- a/students/toddschultz/session01/comprehensions.py
+++ b/students/toddschultz/session01/comprehensions.py
@@ -1,0 +1,24 @@
+import pandas as pd
+music = pd.read_csv("featuresdf.csv")
+
+def both_gen():
+    ''' for any row, if danceability and loudness pass their tests
+    yield the danceability, artist, and song of that row'''
+    count = 0
+    while count < len(music.danceability):
+        if music.danceability[count] > 0.8 and music.loudness[count] < -5.0:
+            yield (music.danceability[count], music.artists[count], music.name[count])
+        count += 1
+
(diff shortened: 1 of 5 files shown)


ldconejo:
Hi Todd,

Welcome to the class! I have  added you to the MS Teams group account (we no longer use Slack for this class). 

On your question on the iterator, I modified your _iterator_2.py_ to break at i=16 and then resume, then I do something similar, using the _range()_ function instead. See the behaviours and think of how you could make _iterator_2()_ behave like _range_:

`def iterator_2(start, stop, step=1):
	while start < stop:
		yield start
		start = start+step

print("First, using our iterator")
qwerty = iterator_2(10, 30, 3)
for i in qwerty:
    print(i)
    if i == 16:
        break

print("This is after the break, with iterator")

for i in qwerty:
	print(i)

print("Now, using range")
qwerty = range(10,30, 3)
for i in qwerty:
    print(i)
    if i == 16:
        break

print("This is after the break, with range")

for i in qwerty:
	print(i)`

I will close this pull request, so that you can make any changes and submit a new pull request once those changes are in.

Thanks,

Luis

PR #371, opened 2018-11-02
lesson03 complete

seaneightysix:


ldconejo:
Hi Sean,

Good job on both assignments. For this lesson, you need to do separate submissions for the locke and recursion problems, you only did it for recursion, so please go ahead and submit for the locke problem on the course's website.

Thanks,

Luis

PR #379, opened 2018-11-05
Lesson 6 files

Swt16:
Lesson 6 files


ldconejo:
Hi Shin,

Good job on the calculator project. On the water regulation one, everything works fine, but there is one warning on checking with Pylint that should be addressed:

`waterregulation/decider.py:63:16: W0105: String statement has no effect (pointless-string-statement)
`
Should be quick fix.

Thanks,

Luis

PR #380, opened 2018-11-05
Iterator_2 v2

toddlikesbeer:
Luis helped get me unstuck, this is my result. Thanks!


ldconejo:
Hi Todd,

Everything looks good. For assignment #2, remember to submit on the course's website as well, so that I can grade it. The closure you implemented indeed works as one, just keep in mind that you can also pass parameters both to the closure as well as the enclosing function (for example, to create functions that output songs with different danceability levels).

Happy coding!

Luis

PR #392, opened 2018-11-11
adding mailroom db work lesson08

stefanjp1:


ldconejo:
Hi Stefan,

For this one, please enable capabilities to list existing donors / donations as well as to update data (could be a donation's amount or a donor's name, for example).

Thanks,

Luis

PR #396, opened 2018-11-12
Peewee Insert into single column table, when column is Primary Key returns error

wpucilow:
Peewee Insert into single column table, when column is Primary Key returns the following error:
peewee.OperationalError: near "WHERE": syntax error
example:
class Donor(BaseModel):
     donor_name = CharField(primary_key = True, max_length = 30)
database.create_tables([Donor])
new_donor = Donor.create(donor_name = "John Novak")
new_donor.save()
peewee.OperationalError: near "WHERE": syntax error
However after removing "primary_key = True" from column definition, there is no this error. Problem happens in peewee, but not in sqlite3. Is there any W/A available?


ldconejo:
Hi Wieslaw,

Peewee doesn't work well when you try to create a single column table in which that single column acts as a primary key. If you want to make sure _donor_name_ is a non-repeatable value, you can use _unique=True_ instead:

```
class Donor(BaseModel):
    donor_name = CharField(unique = True, max_length= 30)
```

This will have the effect of creating a table with two columns, _ID_, which is populated automatically and set as the primary key, and _donor_name_. 

Another alternative is to simply add another column to the same table:

```
class Donor(BaseModel):
    donor_name = CharField(primary_key = True, max_length= 30)
    donor_address = CharField(max_length= 100)
```
With this, _donor_name_ can be the primary key (just in case, the name of person is usually not a good primary key).

One (dirty) trick you could try, if you were set on having a table with a single column set as primary key, it to catch the exception when invoking the _save_ method:

```
try:
    new_donor.save()
except:
    pass
```
This will result in a database file that will indeed have a single column and that column WILL be the primary key.

This was a very interesting question, as I had not previously seen this scenario (it's uncommon to have a table with a single column that is also the primary key).

Regards,

Luis

wpucilow:
Hi Idconejo,

Thank for your response. I will go for the second column, which can be just empty but will prevent from this issue. Apart from that I was just thinking if we can create a VIEW in peewee. Basically table Donor will have a donor name, table donation will have two columns donor and donation. As for the donor donations number, total and average these values should not be keep in table (in my opinion) but should be calculated under the request, based on the data in donation table. Number/total/average should not be explicitly inserted into table, because it can lead to data integrity violation.
So whatever we can create a VIEW based on Donation table:

CREATE VIEW Report AS

select donor_id, count(amount) as num, sum(amount) as total,  avg(amount) as avg from donation group by donor_id;

PR #399, opened 2018-11-14
report() - type object 'Donation' has no attribute 'select'

wpucilow:


wpucilow:
question: report() fails from run_mailroom_db.py 

report() function works when run from create_donors_db.py
but this same function fails with 

INFO:__main__:type object 'Donation' has no attribute 'select'

when run from main program - run_mailroom_db.py. Did I mess with access to database from main program?

wpucilow:
Please ignore this question. I have found my problem. Basically in main program I have defined class Donation with static methods for projection and multiplication used for challenge (multiplying donations). I have overwritten "name space" thus further when referring to Donation was referral to this class and not to the Donation model. After changing this class name to Projection, it works. My fault and lesson learn about the importance of naming convention. If class Name(BaseModel) is defined and refered in the program. This program should not use "Name" for any other object definition.

Wieslaw

PR #400, opened 2018-11-14
Lesson 1 - 3, 5 Pull request

willhwhite:
Submitting the assignments for lessons 1, 2, 3, & 5 for review

### DIFF
--- a/students/wwhite/Lesson01/generator_solution.py
+++ b/students/wwhite/Lesson01/generator_solution.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python
+
+import math
+
+def intsum():
+    """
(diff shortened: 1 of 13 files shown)


ldconejo:
Hi Will,

For lesson 3, you need to submit separately on the course website (no problem with submitting multiple assignments as a single pull request), so that I can grade it. Your implementation is good, I'd encourage you to try a bit more on the handling of the exception of too many boats by letting the ___exit___ method deal with it:

```
        if exc_type is ValueError:
            print(exc_val) 
        return True
```

Regards,

Luis

willhwhite:
Thanks Luis, I’ll be sure to do this today

PR #402, opened 2018-11-16
adding lesson 09 assignment

TinaBrownfield:
submitting lesson 09 assignment

### DIFF
--- a/students/TinaB/lesson09/threading_get_news.py
+++ b/students/TinaB/lesson09/threading_get_news.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python3
+"""
+Lesson 09
+In this lesson, we looked at an asyncio based web data scraper.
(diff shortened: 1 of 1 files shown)


ldconejo:
Hi Tina,

Your code as it is right now crashes due to the following log entry from line 36:

`logger.info({srcs})`

Now, the _srcs_ variable comes from line 33:

```
srcs = [src['id'].strip() for src in data['sources']]
```
You cannot hash a list, which what causes the script to crash:

```
  File "threading_get_news.py", line 36, in get_sources
    logger.info({srcs})
TypeError: unhashable type: 'list'
```
Was it somehow working on your side?

Thanks,

Luis

TinaBrownfield:
It was working on my end when i submitted it. But now is not. I will edit and resubmit.

PR #419, opened 2018-11-22
adding lesson09 newsapi

stefanjp1:


ldconejo:
Hi Stefan,

Your code is currently returning zero matches, this is due to a very small error in _get_articles()_. In order to find it, I'd recommend the following:

- Do not print the response, it will show a lot of text and make it the whole thing unusable.
- Print instead the value of _resp.status_code_. You should be able to figure out what's wrong.

Also, I'd suggest after you get the sources, try to run only with a subset of those, which will save you from running out of the free tier of the API, you can add this line:

```
sources = get_sources()
sources = sources[:10] # This will cut _sources_ to only the first 10.
```

Regards,

Luis

PR #437, opened 2018-11-28
Lesson08 question for update array

wpucilow:
Hi,

PLease find attached file with mongodb operations that are supposed to be implemented in Mailroom program. I have a questions to step 6 and step 7

Step  6 Projection factor, min=20, max=200

I'm building two cursors (cursor4 and cursor2) and then I merge both using dictionary. Is there any way to combine both cursors like UNION ALL in SQL. How to build complex queries.

Step 7  Apply projected changes: factor, min=20, max=200
I'm looking for a way how to update array elements in collection, but only those which match serach criteria. 
If I'm not wrong $unwind and pipe processing works only for collection.aggregate() - Option 1 and not update.

Either Option 2 - update_many with 
array.$ - updates only single element, 
array.$[] - updates all elements

doesn't let me updates those array elements which match { '$gt' : _min, '$lt' : _max} criteria.

Finally how can I dump collection to json file, the simplest method from script?

### DIFF
--- a/students/Wieslaw_Pucilowski/Lesson08/Assignment2/Update_Array/donor_data.py
+++ b/students/Wieslaw_Pucilowski/Lesson08/Assignment2/Update_Array/donor_data.py
@@ -1,0 +1,39 @@
+"""
+initial donor db dataset
+"""
+__author__ = "Wieslaw Pucilowski"
+
+def get_donor_data():
+    """
+    returns dataset
+    """
+    donor_data = [
+        {
+            'name': {'first_name': 'Brandon', 'last_name': 'Dekinson'},
+            'location': 'Bothell',
+            'donations' : [10, 20, 30]
+        },
+        {
+            'name': {'first_name': 'Susan', 'last_name': 'Vegas'},
+            'location': 'New York',
+            'donations' : [100, 99]
+        },
+        {
+            'name': {'first_name': 'Abdul', 'last_name': 'Alhambra'},
+            'location': 'Damascus',
+            'donations' : [999]
+        },
+        {
+            'name': {'first_name': 'Ivan', 'last_name': 'Smirnoff'},
+            'location': 'Moscow',
+            'donations' : [100, 200]
+        },
+        {
+            'name': {'first_name': 'Speedy', 'last_name': 'Gonzales'},
+            'location': 'Los Angeles',
+            'donations' : [97, 45]
+        }
+    
+    ]
+
+    return donor_data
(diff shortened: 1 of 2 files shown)


wpucilow:
I have also push example code and question for Redis. 
students/Wieslaw_Pucilowski/Lesson08/Assignment2/Redis_List/mailroom_exercise_redis.py

I don't know how to integrate list with donor's hash. Basically donors information like name, zip, phone are storred in hash data type. Donations are in list. But from the info I have found Redis does not allow for hash of list. I'm not sure what should be the bet solution here.

On the other hand I would like to make sure what is the scope of these 3 assignment. What is expected from redis and neo4j part. I'm not going to refractor entire code to have every single feature working, basically it doesn't make sense to repeat the same code in python as it cannot be efficiently done in Database language. I will try to cover all features in mongodb, then for redis and neo4j I will implement report, add, delete user skipping projection/challenge part. 
Please let me kno wif this is OK.

ldconejo:
Hi Wieslaw,

I'll take a look at your pull request and get back to you with my feedback.

Regards,

Luis

ldconejo:
Hi Wieslaw,

For your question about Redis, have you tried using Pickle? As the name more or less suggests, it allows you to _pack_ structures that Redis doesn't necessarily recognize into objects that can be stored it. Here is an example I took from this link:

https://stackoverflow.com/questions/15219858/how-to-store-a-complex-object-in-redis-using-redis-py 

```
import pickle
import redis

r = redis.StrictRedis(host='localhost', port=6379, db=0)
obj = ExampleObject()
pickled_object = pickle.dumps(obj)
r.set('some_key', pickled_object)
unpacked_object = pickle.loads(r.get('some_key'))
obj == unpacked_object
```

Applied to your case, _obj_ would be your hash of list, once "pickled" into the _pickled_object_ variable, you can do r.set and r.get on it, and use pickle.load to recover the original structure.

-Luis

ldconejo:
For UNION ALL, try the _setUnion_ command, here's the syntax:

https://docs.mongodb.com/manual/reference/operator/aggregation/setUnion/ 

That should return a combined collection of elements from both cursors.

ldconejo:
For updating array elements, you might want to try _arrayfilters_. Here's the documentation for MongoDB on the topic:

https://docs.mongodb.com/master/release-notes/3.6/#arrayfilters

ldconejo:
Finally, for exporting a collection to JSON, you should use _mongoexport_:

https://docs.mongodb.com/manual/reference/program/mongoexport/ 

That one will let you create either a JSON file or a CSV file.

Regards,

Luis

PR #442, opened 2018-12-01
Lesson 10, submit

mcbraunie:
RE: Play with Cython and the Great Circle code >> complete


ldconejo:
Hi Micah,

On the assignment for lesson 9, you could do the following:

- For debug purposes, you can limit the number of news sources you're polling from, that way you won't run out of the API free tier. You can do it like this:

```
for source in sources[:5]: # This will limit your polling to only the first 5 news sources
```
- Import _Semaphore_ from _threading_, you will need that to ensure the threads don't trample each other accessing the _titles_ list:

```
from threading import Thread, Semaphore
```
- Modify the _get_articles()_ method, so that it takes a source, the _titles_ list and the semaphore. It should look like this:

```
def get_articles(source, titles, sem):
    """

    """
    url = base_url + "articles"
    params = {
        "source": source,
        "apiKey": NEWS_API_KEY,
        "sortBy": "top",
    }
    print('Requesting: ', source)
    resp = requests.get(url, params=params)
    if resp.status_code != 200:
        print('An error occurred with {}'.format(source))
        print(resp)
        print(resp.text)
        return []
    data = resp.json()
    print('Retrieved: {}'.format(source))
    sem.acquire()
    titles.extend([str(art['title']) + str(art['description']) for art in data['articles']])
    sem.release()
    return titles
```

- Finally, modify the _main_ block:

```
if __name__ == '__main__':
    start = time.time()
    get_sources(sources)

    sem = Semaphore()

    list_items = []


    for source in sources[:5]:
        thread = Thread(target=get_articles, args=(source, titles, sem))
        thread.start()
        list_items.append(thread)

    for thread in list_items:
        thread.join()

    print('Process took {} seconds'.format(time.time() - start))
    print(titles)
```

Note that in this case, I'm only printing the list of titles provided by the threads, you would still need to do the word count function, which should be relatively easy to implement.

I will grade the assignment, but please feel free to submit an updated version with the word count implemented and I'll update the grade.

-Luis

mcbraunie:
Thank you, instructor -

I have made the edits per your suggestions and submitted the pull request.

All best,

Micah Braun

PR #444, opened 2018-12-02
Lesson 9 Assignment  Debug

Swt16:
Hello Luis,

I'm having some minor difficulties with this assignment.
1. I'm not sure how to include more threads in the code.
2. It takes a long time to run, equally as long as the get_news_sync program, since it was basically copied from there.
3. To be honest, I don't think I understand threading very well.

Thanks,
Shin

### DIFF
--- a/students/ShinTran/lesson09/get_news_threading.py
+++ b/students/ShinTran/lesson09/get_news_threading.py
@@ -1,0 +1,102 @@
+"""
+Shin Tran
+Python 220
+Assignment 9
+
+Regular synchronous script to see how much a given word is mentioned in the
+news today
+
+Uses data from the NewsAPI:
+https://newsapi.org
+
+#!/usr/bin/env python
+"""
+
+import queue
+import requests
+import threading
+import time
+
+WORD = "trump"
+NEWS_API_KEY = "7acedf3749254086b464c43945e071a4"
+base_url = 'https://newsapi.org/v1/'
+
+
+def get_sources():
+    """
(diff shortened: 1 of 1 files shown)


ldconejo:
Hi Shin,

You're not that far from having it working. Some things you can change:

1. Reduce the number of sources you're polling from. This is just to avoid running out of the free tier on the API. For example, this will limit the run to only the first five sources.

```
    for source in sources[:5]:
```

2. Change _get_articles()_ to take _source_, _titles_ and _queueLock_. _queueLock_ will make sure that only thread is adding articles to the _titles_ list. Updating _titles_ within _get_articles()_ should look like this:

```
    queueLock.acquire()
    titles.extend([str(art['title']) + str(art['description'])
              for art in data['articles']])
    queueLock.release()
```

Note that you're now _extending_ the list with the articles from each particular source, so that you get a combined list in the end. The lock forces other threads to wait for the thread holding the lock before adding their respective articles. This improves execution, as one thread can be updating the list while the others are polling websites for information.

3. Within the main block, starting your threads will look like this:

```
    titles = []
    for source in sources[:5]:
        thread = threading.Thread(target = get_articles, args = (source, titles, queueLock))
        thread.daemon = True  # allow ctrl-c to end
        thread.start()
        thread_list.append(thread)

    for j in thread_list:
        j.join()

```
4. You will need to implement your word count after this, on the final list.

Let me know if this helps or if you have any other questions.

Regards,

Luis

PR #446, opened 2018-12-03
Lesson09 - race conditions

wpucilow:
Hi Luis,

I have started lesson 09 and having first question for this one. I have uploaded modified race_conditions.py. First my understanding is:

func() update class variable val. In for loop we're running 100 threads, each one handling func(). Since All of the threads are accessing the same class variable, some of them can't write (increament) its value.
However In my test I have noticed that the time.sleep() is smaller the more shared.val  is incremented.
What does it mean it fact? Why there is less conflicts for the same 100 threads if the func() takes only .000001 sec then the same number of threads when func() takes 0.01 sec. All of them are trying to access the same shared.val

test 2

I have updated class shared with method func() which increment shared.val(), creating an object
shrek = shared(1), and then in another for loop I'm running 100 threads for shrek.func()
thread = threading.Thread(target=shrek.func). when comparing both (class val access from external function, and class val access from an object) the second works well. example for time.sleep(0.0001

+++ Class val
28
+++ Class Object val,
100

Why there are no conflicting threads when incrementing value from object method, than from external method.

class shared:
    val = 1
    val2 = 1
 
    def __init__(self, y):
        self.val = y
    
    def func(self):
        time.sleep(0.0001) # self.val = 100
        self.val +=1
        shared.val2 += 1

### DIFF
--- a/students/Wieslaw_Pucilowski/Lesson09/integrate_threads.py
+++ b/students/Wieslaw_Pucilowski/Lesson09/integrate_threads.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python3
+
+import threading
+import queue
+
+# from integrate.integrate import integrate, f
+from integrate import f, integrate_numpy as integrate
+from decorators import timer
+
+
+@timer
+def threading_integrate(f, a, b, N, thread_count=2):
+    """break work into N chunks"""
+    N_chunk = int(float(N) / thread_count)
+    dx = float(b - a) / thread_count
+
+    results = queue.Queue()
+
+    def worker(*args):
+        results.put(integrate(*args))
+
+    for i in range(thread_count):
+        x0 = dx * i
+        x1 = x0 + dx
+        thread = threading.Thread(target=worker, args=(f, x0, x1, N_chunk))
+        thread.start()
+        print("Thread %s started" % thread.name)
+
+    return sum((results.get() for i in range(thread_count)))
+
+
+if __name__ == "__main__":
+
+    # parameters of the integration
+    a = 0.0
+    b = 10.0
+    N = 10**8
+    thread_count = 1
+
+    print("Numerical solution with N=%(N)d : %(x)f" %
+          {'N': N, 'x': threading_integrate(f, a, b, N, thread_count=thread_count)})
+
--- a/students/Wieslaw_Pucilowski/Lesson09/race_condition.py
+++ b/students/Wieslaw_Pucilowski/Lesson09/race_condition.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python3
+
+import threading
+import time
+
+
+# create a mutable object that is shared among threads
+class shared:
+    val = 1
+    
+    def __init__(self, y):
+        self.val = y
+    
+    def func(self):
+        # time.sleep(0.1) # self.val = 1
+        # time.sleep(0.01) # self.val = 90
+        # time.sleep(0.001) # self.val = 95
+        time.sleep(0.0001) # self.val = 100
+        # time.sleep(0.00001) # self.val = 100
+        self.val +=1
+
+
+    
+shrek = shared(1)
+
+def func():
+    y = shared.val
+    # time.sleep(0.1) # shared.val = 2
+    # time.sleep(0.01) # shared.val = 8
+    # time.sleep(0.001) # shared.val = 18
+    time.sleep(0.0001) # shared.val = 28-32
+    # time.sleep(0.00001) # shared.val = 28-33
+    # time.sleep(0.000001) # shared.val = 58-97
+    # time.sleep(0.0000001) # shared.val = 64-98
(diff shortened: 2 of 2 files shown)


wpucilow:
BTW with lock increment count improves from 28 to  80-90

lock = threading.Lock()

def func():
    lock.acquire()
    y = shared.val
    time.sleep(0.0001) # shared.val = 28-32
    y += 1
    shared.val = y
    lock.release()

ldconejo:
Hi Wieslaw,

Let me take a look and get back to you.

-Luis

ldconejo:
Hi Wieslaw,

My apologies for the delayed reply to this one. This week has been comparatively packed with assignment submissions. Nevertheless, here are my answers:

_Since All of the threads are accessing the same class variable, some of them can't write (increment) its value._

Not exactly. The first thing here is that you should have a _thread.join()_ after all your loops have been launched:

```
for i in range(100):
    thread = threading.Thread(target=func)
    threads.append(thread)
    thread.start()

thread.join()
```

Otherwise, some of your loops may still be running by the time your print statements at the end run.

Beyond that, it's that not some of the threads will not be able to increment the class variable, they all get to write the variable in turn, the thing here is the initial value of _shared.val_ they get. Let's imagine an extreme case with only two threads and a very long sleep (2 seconds).

When the first thread is launched, the first thing it does is to make a copy of the value of shared.val:

# Thread 1
y = shared.val # shared.val = 1

After that, it sits there for there for two seconds. Meanwhile, the second thread is launched and does the same, only a fraction of a second later:

# Thread 2
y = shared.val # shared.val = 1

The two seconds have passed, now thread 1 does this:

y += 1 # y = 2
shared.val = y # shared.val = 2

Then thread 2 also comes back and does exactly the same thing, since it's initial value for y was also 1.

Now, as you make the sleep smaller and smaller or even remove it, this happens:

```
# Thread 1:
y = shared.val # shared.val = 1
time.sleep(0.00000000000000001) # If this sleep is small enough, it might be over before Thread 2 is launched
shared.val = y # shared.val = 2
```

Now, if thread 2 launches AFTER thread 1 has already finished, this will happen:

```
y = shared.val # shared.val = 2, thread 2 already got the updated value of y
```
With 100 threads, the smaller you make the sleep, the more you increase the chances of threads finishing and updating before new threads come into existence.

In this case, for a lock to be effective, you'd need to "guard" the entire function, forcing the threads to run sequentially:

```
def func():
    lock.acquire()
    y = shared.val 
    # No need to sleep here, the lock prevents other threads from accessing this code at the same time
    y += 1
    shared.val = y
    lock.release() 
```
You could also get the same result (a sequential run of the threads) by doing this:

```
for i in range(100):
    thread = threading.Thread(target=func)
    threads.append(thread)
    thread.start()
    thread.join()
```

This would not be a good idea, as the join forces the thread to finish before the next thread can start, so the advantage of threading gets lost. Locks are useful, though, when your threads need to complete other tasks that do not involve shared resources and only sporadically access one or two shared variables which you protect with the lock.

To your other question:

_Why there are no conflicting threads when incrementing value from object method, than from external method._

The _func_ class method you have is different, as you modify the values of _self.val_ and _shared.val2_ directly. The GIL (Global Interpreter Lock) will NOT allow multiple threads to modify a memory location (i.e., where the values those variables are stored) at the same time, so they all have to wait for their turn. Try changing _func_ within the class to something like this:

    def func(self):
        time.sleep(0.01)
        y = self.val
        y += 1
        self.val = y
        shared.val2 += 1

Each thread will get their own "y" variable, which the GIL allows them to modify freely. They only have to get in line to write that value to self.val.

Let me know if this makes sense.

-Luis

wpucilow:
Thanks Luis,

Two good points. I can see now what the race conditions mean. As for the second point I can see why we should assign shared memory to the local variable, which then will be incremented. Each of the thread will keep the local copy of the variable for incrementing. However my question was for the following

class shared:
    val = 1
    val2 = 1
    
    def __init__(self, y):
        self.val = y
    
    def func(self):
        time.sleep(0.0001) # self.val = 100
        y = self.val    # each thread will make a copy of shared val to its own y variable
        y += 1          
        self.val = y
        z = shared.val2
        z += 1  
        shared.val2 = z


shrek = shared(1)

# external func()
def func():
    y = shared.val
    time.sleep(0.0001) 
    y += 1
    shared.val = y
 
1.) updating the class variable with external func()

for i in range(100):
    thread = threading.Thread(target=func)
    threads.append(thread)
    thread.start()

+++ Class val
14

2.  Updating shrek = shared(1) object variable using object method

for i in range(100):
    thread = threading.Thread(target=shrek.func)
    threads.append(thread)
    thread.start()

+++ Class Object val, 
97
100

with the same condition 0.0001 sec, when updating shrek.val object variable using object method shrek.func() we're getting quite good result = 97, when updating class shred.val with external func() we are getting 14.

PR #454, opened 2018-12-05
Trouble with Lesson08 mr_neo4j.py 

seaneightysix:
Hi,

I am having trouble with my mr_neo4j.py script and would like assistance please. 

It seems like the following context manager only works on lines 15 and 19:
driver = mr_login_database.login_neo4j_cloud()
with driver.session() as session: 

Whenever I use the above context manager in any of my methods, it does not work: the mr_login_database.login_neo4j_cloud() returns an empty driver.

Your assistance is greatly appreciated.

Thanks,
Sean


ldconejo:
Hi Sean,

I think the issue might be related to a typo you have when iterating over the results. Using the _list_donors()_ method as example

```
def list_donors():
    log.info("ENTERING LIST DONORS")
    driver = mr_login_database.login_neo4j_cloud()
    with driver.session() as session:
        cyph = """
                MATCH (person:Person)
                return person
                """
        results = session.run(cyph)
        all_donors = []
        for donor in results:
            all_donors.append(donor['Person']['Name'])
    return all_donors
```

Your _cyph_ returns _person_ (all lowercase) not _Person_, so line 119 (right before _return all_donors_) should be:

```
all_donors.append(donor['person']['Name'])
````

With that change, I get the driver session within that method to return the list of existing donors.

Let me know if that solves your issue.

Thanks,

Luis

PR #463, opened 2018-12-08
Threads/coroutine/yield

wpucilow:
Hi Luis,

I have the following subject I would like to discuss. I have uploaded 3 files:
- video2_fibonacci.py - fiboinacci with basic task loop and coroutine,
- video3_web_downloader.py - sequencial program to download and grep key word in news, mind to update NEWS_API_KEY = "<put your key here>"
- video3_web_download_threads.py - program as above but using the threads (not verified)

I have the following question:

1. multithreading:
video3_web_download_threads.py, I'm running one thread per source. In one thread I'm doing get_articles(source), count_word, etc. Is it enough to run each get_articles(source),
 which is request.get(url) in separate threads, then the OS itself makes sure to minimize the I/O waiting time> In other words how to run multiple url request in the threads to optimize 
the overal script running time.

def check_word(source):
    titles = get_articles(source)
    global art_count
    global word_count
    art_count += len(titles)
    word_count += count_word('trump', titles)

2. coroutine:
I would like to run the code from video3_web_downloader.py but using basics TaskLoop and coroutine/yield as it is done in video2_fibonacci.py. I don't want to use asyncio framework 
which does this job. Instead I would like to use the solution as in video2_fibonacci:
task in the even loop is fib() function which await sleep(), if sleep() is not completed yield is send to the TaskLoop. This makes taks to be moved from the top of the "stact" to the very end, 
and another task is processed. If fib()-sleep() is completed then StopIteration exception is capture in the TaskLoop, and fib() result is passed to it. 

How to achive similar thing with request.get(url) ? 
More or less I can imagine that get_source() must be completed first, no coroutine here. then in TaskLoop tasks could be count_word(word, titles, source) so there are as many tasks 
as web sources to request.

inside count_word() - await get_article()
inside get_article() - resp = requests.get(url, params=params), 

and here is my question. inside  get_article() we are supposed to run requests.get(url, params=params) and if it is not accomplished we should yield "something" 
to the TaskLoop in order to move forward to another task. How can we achive this. Should some time sleep be introduce. Then what is the yield condition. 
Once we do this in the code: resp = requests.get(url, params=params) we cannot move forward to another task.

### DIFF
--- a/students/Wieslaw_Pucilowski/Lesson09/async/video2_fibonacci.py
+++ b/students/Wieslaw_Pucilowski/Lesson09/async/video2_fibonacci.py
@@ -1,0 +1,80 @@
+from types import coroutine
+import time
+
+@coroutine
+def sleep(secs=0):
+    start = time.time()
+    while time.time() - start < secs:
+        # now we need it to yield control
+        yield "yielding in sleep diff: %s < secs: %s" % (time.time() - start, secs)
+    return "{} seconds have passed".format(time.time() - start)
+
+"""
+Now we'll create a little coroutine that calculates something,
+"""
+async def fib(n):
+    """
+    Classic fibbonacci number, but with a delay
+    """
+    if n == 0:
+        return 0
+    a, b = 0, 1
+    for i in range(n-1):
+        a, b = b, a + b
+    await sleep(0.001) # - is calling awake on sleep before it actually returns the answer,
+    return b
+
+class TaskLoop():
+
+    def __init__(self):
+        # lsit to hold the tasks
+        self.tasks = []
+
+    def add_task(self, task):
+        """
+        add a task to the loop task must be a coroutine
+        """
+        self.tasks.append(task)
+
+    def run_all(self):
+        """
+        This is where the task loop runs
+        """
+        results = []
+        # Keep a loop going until all the tasks are gone:
+        i = 0
+        while self.tasks:
+            i += 1
+            time.sleep(0.0)
+            print(f"\nOuter loop count: {i}")
+            # pop a task off the end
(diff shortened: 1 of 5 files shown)


wpucilow:
Generally speaking point 2.)  is about how to create coroutine/generator which start resp = request.get(url) and while request is not completed, it yields something to the TaskLoop in order to switch to the next task in the list, instead of waiting.

ldconejo:
Hi Wieslaw,

For the first question, yes, placing the requests to each news source on a separate thread is a good solution. I would suggest you use the _queue_ module to create a queue for each thread to place the list of titles returned by their corresponding news source. After _thread.join()_ you can use another function to go over the queue counting the words (you could also do the count at thread level and just place an integer in the queue).

Regards,

Luis

ldconejo:
For the second question, there's an easier way to implement it with coroutines than the Fibonacci example. First, you get a list of news sources, the coroutine function will be in charge of polling each source for new titles. 

get_headlines_coroutine():
    while True:
        source = (yield)
        print("Getting news from: ", source)
        # add code here to get the headlines from the source and put the results somewhere (maybe a queue?)

sources = get_news_sources()

poll_sources = get_headlines_coroutine()
# This is to start the coroutine (almost literally, "booting" it)
next(poll_sources)

# Go over the list of sources, sending one source at a time 
for source in sources:
    poll_sources.send(source)

Let me know if this makes sense.

Regards,

Luis

ldconejo:
Another tip for this assignment: Try to reduce the list of news sources you poll, otherwise you'll run out of the free usage tier of the API, do polling on five sources or so each time (just do sources = sources[:5]).

-Luis

PR #465, opened 2018-12-09
Assignment 8 from Cindy Waldron

cindywaldron:


ldconejo:
Hi Cindy,

For the Redis portion of the assignment, could implement a form of data lookup? As suggested, you can for example allow search for the email address of a donor. For MongoDB, please enable the capability to update the data (for example, the name of a donor or the amount of an existing donation. 

Thanks,

Luis

PR #467, opened 2018-12-10
Lesson09 assignment - news scanner threads

wpucilow:


ldconejo:
Hi Wieslaw,

For Hadoop, start with their official documentation here:

https://hadoop.apache.org/ 

There's a "Getting started" section with information on how to setup a single-node cluster. Big data is a whole field on its own, but if might be worth the effort if you will be regularly dealing with large amounts of data like in the example you proposed.

The same goes for Spark, go to their official page:

https://spark.apache.org/ 

And go to the "Getting started" section. Take a good look at the Wikipedia pages I provided before, so that you more or less know what each one does.

Regards,

Luis

PR #482, opened 2018-12-14
Lesson09 multithread download assignment

willhwhite:
I accidentally used an old comment when I originally committed, so I recommitted with a new comment. Please disregard the message about Lesson07


willhwhite:
Thanks Luis for the feedback, will look into that now

PR #483, opened 2018-12-15
Resubmission of Lesson07 and Submission of Lesson10

seaneightysix:
Added the function pretty_print_query() which "produce(s) a list using pretty print that shows all of the departments a person worked in for every job they ever had".


seaneightysix:
Hi, along with resubmitting the required changes for Lesson07, I've also included my submission for Lesson10.

PR #486, opened 2018-12-16
lesson 10 assignment

cosmonautz:


ldconejo:
Hi Kevin,

Could you submit the package containing the _profile_ decorator you're using as part of the pull request?

Thanks,

Luis

cosmonautz:
I just used the memory profiler with the command  "python -m
memory_profiler trigrams.py A_Connecticut_Yankee_in_King_Arthurs_Court.txt
output.txt", is that what you are looking for?

ldconejo:
Hi Kevin,

Ok, that makes sense. Thanks for the clarification.

Regards,

Luis

PR #490, opened 2018-12-17
Pull request for Lessons08/09/10 assignments

willhwhite:


willhwhite:
I'm having difficulty pushing the nosql file from Lesson08. I assume it is because this repo was cloned, but cannot figure out how to push my updates to github. Do you know what the issue might be?

ldconejo:
Hi Will, 

Do you get any error messages for the files on lesson 8?

Thanks,

Luis

willhwhite:
I didn't receive any errors. I usually add, commit, & push through Pycharm, but for some reason Pycharm seems to be completely ignoring these files. I tried manually adding the files then pushing to git through Git Bash, but it seems to be pointed at Gitlab instead of Github.

ldconejo:
Which files are being left out? One thing to check would be you .gitignore file.

willhwhite:
All the files from the cloned repo are being ignored. I'm not having issues with other files, as I was just recently able to add/commit/push the files I created for lesson 10

ldconejo:
Interesting issue. Try putting the files inside of the lesson 10 folder (inside of their own subfolder, of course), then do git add, commit and push. That should at least allow me to look at your code.

Luis

willhwhite:
Ok I was finally able to do that this morning after trying and failing last night. It looks like the only files I was able to get github to recognize were those within the src file (I changed to 'source'), so I think some of it may break when looking for the files in the in the 'data' folder

willhwhite:
Also, it looks like lesson 9 was graded and given a 4/8. Will this grade be updated now that the word search has been added to the updated file in this pull request?

ldconejo:
Hi Will,

For lesson 8, the first part is good. You didn't enter a submission for the second assignment of that lesson. Please go ahead and do that, so that I can grade it.

For lesson 9, there are multiple changes that you need to make:

1. You need to do a .join() on each thread once you have launched all of them. This will prevent the word count from starting before all threads have returned. Something like this will do the trick (btw, change the name of the _list_ variable to something else, that's a reserved word in Python and can cause issues):

```
    for thread in list:
        thread.join()
```

2. Do not use the variable _titles_ as a global, you're extending it every time, so that each group of titles added to the queue is a copy of all titles from threads that have already returned, plus the new set of titles. Instead, create _titles_ as an empty list within _get_articles_ and then change your comprehension to do this (instead of extending _titles_):

```
titles = [str(art['title']) + str(art['description']) for art in data['articles']]
```

Then you can add the new titles to the queue, as you are doing already.

One final recommendation: Don't go through the entire list of sources, that will run you out of uses of the free API key. Instead, after getting your initial list of sources, do this:

```
sources = sources[:5]
```
That way, you only work with the first five sources (more than enough).

-Luis

ldconejo:
To your question on regrading assignment 9, yes, I always update the grade once a student has submitted requested fixes.

-Luis

PR #499, opened 2018-12-21
HW 10 from Cindy Waldron

cindywaldron:
Memory Profiler is used for this assignment. for test results, see output.txt.  for code, see mailroom.py

### DIFF
--- a/students/cindywaldron/hw10/create_donor.py
+++ b/students/cindywaldron/hw10/create_donor.py
@@ -1,0 +1,70 @@
+
+import logging
+from peewee import *
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+database = SqliteDatabase('donation.db')
+database.connect()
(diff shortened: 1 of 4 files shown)


ldconejo:
Hi Cindy,

Were you able to draw any conclusions about your code from the memory profiler results? From _output.txt_, memory usage is almost unchanged through the execution of the script.

Thanks,

Luis

cindywaldron:
Hi Luis,

I don't think I can draw a conclusion on my profiler results because my
program doesn't use much memory.  All I can say it uses 32 MB of memory.
Should I run profiler against a different program?

Thanks.

ldconejo:
Hi Cindy,

I think it might be more interesting for you to try to profile code that appears to do the same but behaves quite different in memory. One very simple option is to compare memory usage between lists and generators. 

Luis

cindywaldron:
Hi Luis,

I used a different python program to show more memory usage. Here is the
pull request:

https://github.com/UWPCE-PythonCert-ClassRepos/SP_Online_Course2_2018/pull/503

thanks.

PR #501, opened 2018-12-25
Lesson 06 submission

andykhkwok:
Luis, 
Lesson 06 is ready to be reviewed.  Lesson 05 part one has also been updated.

A few notes, I had to remove one of the files in my local branch for another student because it had an ADD/ADD error as I am trying pull from upstream. In doing so, I have also tested a few git features (rebase, decomple...etc) because I was testing out various functions and it has accidentally deleted my original code. I tried to point back to when I did my previous submission. Please let me if I should restart everything again (deleting existing folk and start over again) then add lesson 06 for re-submission.

For the assignments, I have to add some dummy methods to pass the pylint evaluation because it would say that my classes should have more methods. I also have to rename some files to get the tests to pass due to character issues. 

For Assignment 05 - 01, debugging txt has been updated with additional comments regarding the debugging process. It appears that I cannot do a resubmit in the course website once an initial submission has been done. Please advise if that needs to be updated as well.


ldconejo:
Hi Andy,

Sounds you had quite a bit of fun with this pull request, as well as with the assignments. For lesson 5, your debug log is fine, I have updated your grade. For lesson 6, both assignments are good. I usually ignore Pylint's warnings about too few methods, but your solution is equally valid for assignment purposes. Some development teams will modify Pylint's configuration file according to their needs, so that they are not forced to deal with rules that are not important to them. Great job on these assignments!

-Luis

PR #503, opened 2018-12-26
HW 10 from Cindy Waldron

cindywaldron:
I switched to a different python program to show more interesting memory usage.


cindywaldron:
Hi Luis,

I finished 2nd course, but I haven't heard back.  Could you check?

Thanks.

ldconejo:
Hi Cindy,

Let me check with the Academic Coordinator. I entered your passing grade on December 26th, so you should have received confirmation already.

Regards,

Luis

PR #504, opened 2018-12-27
Finished Lesson 7 assignment 1

josh1209:
Finished the person_job_department assignment


ldconejo:
Hi Josh,

The assignment looks good, the implementation of the database is correct. The one correction you need to make is that, when pprinting the list of jobs, the goal, is to show the department's name, not the code, so that you combine data from the _Job_ and _Department_ tables.

Thanks,

Luis

josh1209:
@ldconejo Got it, I'll adjust it and tag you when I merge again.

Thank you,

-Josh

PR #537, opened 2019-01-10
correction for HW 1

ZhuojingXie:
thank you for correcting me


ldconejo:
Hi Xie,

For the Locke assignment on lesson 3, add _return True_ at the end of the exit method will prevent your script from crashing:

```
    def __exit__(self, etype, evalue, tb):
        if etype == ValueError:
            print("WARNING: Too many boats through.")
        return True
```
Code does look good and it's working.

ldconejo:
On your earlier question on why the .csv file is not uploaded to Github, check the exclusions in your _.gitignore_ file. That's what prevents your _featuresdf.csv_ from being uploaded.

Regards,

Luis

PR #538, opened 2019-01-10
Lesson 8 Assignment 2

josh1209:
Refactored mailroom to utilize each specified database (pymongo, redis, and neo).


ldconejo:
Hi Josh,

For MongoDB, I ran into the following issue when deleting a donation:

```
Traceback (most recent call last):
  File "mailroom_mongodb.py", line 290, in <module>
    c = switch_func_dict.get(choice, continuing)()
  File "mailroom_mongodb.py", line 184, in delete
    set_variables(query, temp)
  File "mailroom_mongodb.py", line 89, in set_variables
    setting.first_gift
  File "mailroom_mongodb.py", line 66, in first_gift
    {'$set':{'first_gift': temp[0]}}
IndexError: list index out of range
```

This one only happens when a donor has a single donation and that donation gets deleted. 

For the Redis part, one of the requirements is to implement some sort of lookup functionality. I'd recommend one that will return a donor's email (as suggested by the assignment).

Good implementation otherwise, I agree with many of your comments on the advantages and / or difficulties encountered with one database or the other.

Regards,

Luis

PR #539, opened 2019-01-11
Lesson1 - submit1

githubtater:


ldconejo:
Hi Roy,

Generators and list comprehension are working ok. For the iterator, the idea is that you can do a break and the sequence will restart (which would happen if using the _range_ function). The way it is done right now, it will simply resume from where the break occurred.

Thanks,

Luis

PR #542, opened 2019-01-13
lesson02: generators and closures

tlugosan:


ldconejo:
Hi Toni,

Your closure implementation is good. For the generator, what you're creating is a list object, which is different. You can verify that by adding this line after _list3_ is defined:

print(type(list3))


You will get the following:
```
<class 'list'>

```
Another thing you can do to verify that it is indeed not a generator is to try to do _next_ on it:
```
print(next(list3))
```
Which will result in:
```
TypeError: 'list' object is not an iterator
```
Your definition of list3 is currently as follows:
```
list3 = [x for x in zip(music.artists, music.name) if x[0] == "Ed Sheeran"]
```
See what happens if you replace the outermost square brackets with parentheses. The fact that it becomes a generator does make a difference in several applications, for instance, regarding memory usage.

I'll update the grade once you submit a fix.

Regards,

Luis

tlugosan:
Fixed it with https://github.com/UWPCE-PythonCert-ClassRepos/SP_Online_Course2_2018/pull/547

PR #546, opened 2019-01-16
HW lesson04

ZhuojingXie:


ldconejo:
Hi Xie,

The implementation works correctly, however, the idea was to use the json_save module, either as a metaclass or a decorator to convert the data in your existing mailroom program into JSON objects. This requirement needs to be fulfilled in order to get the full grade for this assignment.

Regards,

Luis

ldconejo:
Hi Xie,

For lesson 5, everything looks great. Note that you need to submit on the course website for the logging assignment as well (you only submitted for the debug one), so that I can grade it.

Regards,

Luis

PR #548, opened 2019-01-16
Lesson 7 - Part 1, 2 and Lesson 8 - Part 1 submission

andykhkwok:
All three assignments are ready for review. 
For reference:
Lesson 07 - Part 1 - direct modification based upon the exercise
Lesson 07 - Part 2 - create_mailroom_db.py gives the parameters; main_mainroom_db.py provides initial data and the interface for additional processing
Lesson 08 - Part 1 - learnnosql.py is the file that exemplifies all different database.


ldconejo:
Hi Andy,

For lesson 7 part 1, the implementation of the Department table is good, with the exception that _dept_num_, not _dept_name_ should have been selected as key. This causes that, when printing the department an employee has worked in, you will need to combine data from the Job and Department tables, which is a common occurrence when working with SQL databases.

-Luis

PR #552, opened 2019-01-21
submit lesson 3 assignments

cocokaku:


ldconejo:
Hi Colleen,

For the recursion part, you're using recursion correctly, but the implementation of factorial is incorrect, essentially you're returning the factorial from the next integer. For example:

Your code for 9! returns 362,880, which is the correct value for 10!. It should be an easy fix. Also, you need to submit the recursion assignment separately on the course website for us to be able to grade it. Let me know once the fix and submission are ready.

Thanks,

Luis

cocokaku:
Luis,
Hi again. I have also corrected and resubmitted the context manager part of
the assignment to GitHub:
https://github.com/UWPCE-PythonCert-ClassRepos/SP_Online_Course2_2018/pull/574
and attached it to this email fyi (there didnt appear to be a way to
resubmit to the website). Thanks very much for the feedback, and for the
opportunity to fix it!
Thanks,
Colleen

ldconejo:
Hi Courtney,

My apologies on the recursion assignment. You're totally right. I have also reviewed the context manager changes, everything looks good. Great job.

Regards,

Luis

PR #553, opened 2019-01-22
lesson07 HW

ZhuojingXie:


ldconejo:
Hi Xie,

For the Department assignment, the pprint data should show the department name, instead of its code. The goal is that is for you to combine data from two different tables.

For the Mailroom assignment, everything looks great, you still need to add functionality to update data, which could be updating a donor's name and/or the value of a donation.

Regards,

Luis

PR #559, opened 2019-01-24
Completed Lesson 10 assignment.

Parzival-X:
The assignment included working with the Great Circle code, Cython profiling, adding a Makefile and unit tests. I included the output of the profile tests in a separate text file.


ldconejo:
I'm glad to see that you took advantage of this assignment to pursue more than just one of the options.

-Luis

PR #569, opened 2019-01-31
lesson04 pull

githubtater:
I revamped my code for mailroom from the first quarter.  In the process, I broke the challenge/projection functions.  These are the functions that included map, filter, etc.

That said, the application still functions as expected in all other areas.  Donors can be printed, have donations added, and new donors entered.  The json save/load functionality is working as well.

### DIFF
--- a/students/roy_t/lesson04/json_save_dec.py
+++ b/students/roy_t/lesson04/json_save_dec.py
@@ -1,0 +1,146 @@
+#!/usr/bin/env python
+
+"""
+json_save implemented as a decorator
+"""
+
+import json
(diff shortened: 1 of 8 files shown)


ldconejo:
Hi Roy,

For the debugging assignment in lesson 5, it looks like you debugged the code that is part of the lesson and not the one for the assignment. One of the goals is to show that you can use the Python Debugger to find out what's wrong with the code (is a simple issue, but good for practice). Please do your debug on the assignment's code and use (an attach your log) Python Debugger to solve it.

Thanks,

Luis

PR #571, opened 2019-02-01
HW lesson10

ZhuojingXie:


ZhuojingXie:
Good afternoon, Professor Luis,

I am still waiting for the lesson09 be graded. Another question regarding
pass the class. The instruction says it only needs cumulative 80% for
passing the class. Is that mean I can skip one of the lessons but still
pass the class as soon as the cumulative score is above 80%?

Thank you

Best regards,
Zhuojing Xie

Luis Conejo-Alpizar <notifications@github.com> 于2019年2月2日周六 下午1:06写道：

ldconejo:
Hi Zhuojing,

I graded your submission for lesson 09 a few days ago, it already shows up on your grade. Your grade is currently 79. As you point out correctly, 80% is the passing grade, regardless of whether you have any pending assignments. For assignments in which you missed points for any pending requirements or issues to correct, you can submit an updated pull request with additional changes to recover those points.

Regards,

Luis

PR #573, opened 2019-02-02
Lesson 10 Progress

andykhkwok:
Hi Luis, 
Please let me know if this would be sufficient for submission if needed. I am still going through the database exercise in Lesson 08 part 2 so i want to budget my time accordingly. 
Thanks, 
Andy


ldconejo:
Hi Andy,

Yes, this is good. I appreciate you including a document summarizing your results. In its current iteration, assignment 10 is wide open in terms of what you can do. I think you made good use of the opportunity. I will keep this pull request open while you submit the assignment, so that I can go ahead and grade it and you can continue working on the second part of assignment 8.

Regards,

Luis

PR #575, opened 2019-02-03
Lesson03 Update: updated context manager error handling

Pjurek3:
* put error handling into exit method
* added test specifically for str Type Errors
* added catch for bad inputs which are strings

### DIFF
--- a/students/paulJurek/lesson03/locke.py
+++ b/students/paulJurek/lesson03/locke.py
@@ -20,9 +20,14 @@ class Locke:
         print('starting pumps')
 
     def __exit__(self, *args):
-        print('open door 2')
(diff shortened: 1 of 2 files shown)


ldconejo:
Hi Paul,

This one is closer, but you can make it simpler. Try replacing your _exit()_ method with the following:

```
        if exc_type != None:
            print(f"Exception found: {exc_val}")
        return True
```

When you raise an exception within _move_boats_through()_, the ___exit()___ method called. You can then check the value of _exc_type_ to determine if an exception occurred and handle it. You don't need the _try/except_ block you currently have within ___exit()___

Regards,

Luis

PR #576, opened 2019-02-03
Lesson04: meta mailroom final commit

tlugosan:


ldconejo:
Hi Toni,

Was the code for lesson 4 working correctly? Right now, I get the following error when trying to open _mailroom5_meta.py_:

```
TypeError: list indices must be integers or slices, not str
```

Part of the issue is the included _donor_list_json_file.json_ file, which has one invalid element that I can see: The topmost element is the list of donors, JSON files are made up of key/value pairs, there is no key pointing to this list.

Thanks,

Luis

ldconejo:
For the logging assignment, make sure that you can see the division by zero error reported by _syslogserver.py_ (which should be running while you run _logging_ex.py._ There's an error in your call to _logging.handlers.DatagramHandler()_, the port parameter should be an integer.


-Luis

PR #589, opened 2019-02-09
submitting lesson 1 assignment

jayjonesseattle:
Please review lesson 1 assignment. I welcome any feedback you may have.

Thank you!

### DIFF
--- a/students/TerranceJ/lesson1/assignment1.py
+++ b/students/TerranceJ/lesson1/assignment1.py
@@ -1,0 +1,103 @@
+
+
+def top_five():
+	"""This is the comprehensions assignment"""
+
+	music = pandas.read_csv("featuresdf.csv")
+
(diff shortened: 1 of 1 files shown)


ldconejo:
Hi Terrance,

First of all, welcome to the course! It looks like you're using Python 2.x for your code, the course requires using Python 3.x (I use 3.7.2, 3.6.x should be ok). This will require you to modify your code for Python 3 and resubmit (should be minor changes).

For the Iterator, you will need to modify the implementation so that it behaves the same as the _range()_ function, that is, it will reset itself after a break. For the generator, you need to include the _test_generator.py_ file included in the assignment. Make sure you run it first, using _pytest_ or _unittest_ and make sure all tests are passing (you will need to do some modifications to the file. Once you're done, go ahead and submit a new pull request.

Thanks,

Luis

PR #592, opened 2019-02-11
assignment for lesson04

elnurad:


ldconejo:
Hi Elnura,

Implementation looks good. I can create a new donor and save it to a JSON file. However, after I exit, enter again and reload the file, the donor that was previously added is not loaded. Please take a look at that and submit a new pull request with your updated code.

Regards,

Luis

elnurad:
Hi Luis,

Thank you. I’ll take a look!

Elnura

PR #593, opened 2019-02-12
Lesson 01 Submittal

Percolator787:
Lesson 01 Submittal.  Files included:
- generator_solution.py
- iterator_1.py
- music_comprehension.py
- test_generator.py
- featuresdf.csv


ldconejo:
Hi Dennis,

First of all, welcome to the course! Your implementations for comprehension and generator are working well. For the iterator, when comparing it to the _range()_ function, you need to modify your iterator to match its behavior. The _range()_ function will restart after a break, so you'll need to change your iterator to reset itself after a break. You can submit a new pull request to fix this.

In terms of code readability, other changes you can make:
- Avoid code / comment lines going beyond 100 characters.
- Leave one space around comparison operators: if one_variable <= other_variable
- Avoid using single-character variables, use descriptive names.
- Do not use _else_ after a _return_ statement:

This:

```
if test_condition == "23":
    return True
else:
    return False

```
Is the same as this:

```
if test_condition == "23":
    return True
return False
```

Let me know if you have any questions!

Luis

PR #596, opened 2019-02-13
lesson10 assignment submission - great circle

mluckeroth:
added improvement to great circle with cython:
-v1 is with variables defined using cdef double
-v2 is with numpy replacing the math module functions
-v3 is with single precision float variable definition
-v4 is with libc.math cimport acos, sin, cos replacing numpy functions


mluckeroth:
Hi Luis,
I reread the assignment and I saw that there should be unit testing included.  I've been trying to get pytest to work using the pytest-cython tool: https://pypi.org/project/pytest-cython/ but I have not been able to make any progress in the last couple days.  I'm also having difficulty finding documentation or examples online.  I will submit a new pull request with the test files that I have created, but if you have any recommendations or examples that I can review that would be helpful.

PR #599, opened 2019-02-14
lesson04/lesson05

elnurad:


elnurad:
Fixed lesson04. Added assignments for lesson05.

ldconejo:
Hi Elnura,

Great job on assignment 1. One minor suggestion is to keep consistency between saving and loading. The saving method automatically adds the _.json_ extension to the file, while the loading requires you to explicitly add that extension. Keep up the good work!

-Luis

ldconejo:
For assignment 5, logging, everything looks good, but you need to modify your log files so that have a date-time stamp as part of their name, as that is one of the assignment requirements. Should be an easy fix, please submit an updated pull request once that feature is added.

elnurad:
Thank you, Luis. Will fix it!

PR #600, opened 2019-02-15
lesson10 adding pytest-cython files

mluckeroth:
Known issues with cdef while running $py.test --doctest-cython using the pytest-cython plug-in.  Not clear on how to resolve, please advise.


ldconejo:
Hi Mark,

I get the following _cdef_ syntax error if I try to run your code without compiling:

```
E       cdef double a #, b, theta, c, x, radius
E                 ^
E   SyntaxError: invalid syntax
```
Is that the error that you mention?

I use this command to generate .so files from your .pyx:

```
$ python3 setup.py build_ext --inplace
```
Now, once I have compiled .so files, I get this error on v2 and v3:

```
>   c = np.acos((np.cos(a) * np.cos(b)) + (np.sin(a) * np.sin(b) * np.cos(theta)))
E   AttributeError: module 'numpy' has no attribute 'acos'
```

Note that arccosine method in _numpy_ has a different name from what _math_ uses. I was able to fix that, recompile v2 and v3 and rerun tests to a successful pass. 

Give it a try and submit a new pull request (should be an easy fix) and you'll be done with this one.

Regards,

Luis

PR #601, opened 2019-02-15
Finished lesson06

githubtater:


ldconejo:
Hi Roy,

Code looks good. Note that one of the requirements of the assignment is to be able to run unit testing from one level up from the _waterregulation_ directory:

_python -m unittest waterregulation/test.py and python -m unittest waterregulation/integrationtest.py should have no failures._

The way _test.py_ and _integrationtest.py_ are coded right now, the code won't be able to find controller.py or decider.py. You'll need to fix that, it's a simple change. Once that is fixed, coverage is at 100% and there are no major issues reported by Pylint or Flake8.

-Luis

PR #603, opened 2019-02-17
submit lesson 4 assignment

cocokaku:


cocokaku:
save and load of donor database is occuring automatically when program starts and ends, rather than commanded by user. code added to main module rather than to donors class with save/load methods. can resubmit if this approach doesnt meet assignment criteria.

PR #605, opened 2019-02-18
Lesson06B

Pjurek3:
Completed water regulation.  My coverage was working with 100% coverage up until last set of commits.  Then it gives report that controller and decider were not imported.  I am thinking I may need to reinstall my coverage which is why I am still submitting.

Also, I received 100% coverage due to last test in test.py.  Not sure if this was intent or not for assignment.

PS C:\Users\pjure\Documents\software\python\SP_Online_Course2_2018\students\paulJurek\lesson06\water-regulation> coverag
e run --source=waterregulation/controller.py,waterregulation/decider.py -m unittest waterregulation/test.py
.............
----------------------------------------------------------------------
Ran 13 tests in 0.000s

OK
Coverage.py warning: Module waterregulation/controller.py was never imported. (module-not-imported)
Coverage.py warning: Module waterregulation/decider.py was never imported. (module-not-imported)
Coverage.py warning: No data was collected. (no-data-collected)


ldconejo:
Hi Paul,

Your code is good, for your coverage command, use _--include=_ instead of _--source==_. The former is used when you want to do coverage analysis for individual files, while the latter (used in the _calculator_ assignment) does analysis for entire folders.

-Luis

PR #607, opened 2019-02-19
Lesson 03 Submittal

Percolator787:
Submitting the following files for lesson03:
- locke.py
- recursion_factorial.py

Also, resubmitted iterator_1.py from lesson01 to fix the IterateMe_2 function to make it act like a range if a break is inserted.


ldconejo:
Hi Dennis,

For lesson 3, you forgot to enter your submission for the Recursion assignment (it's separated from the Context Manager on). Please go ahead and shoot me an email when you've added that submission, so that I can grade it.

Regards,

Luis

PR #608, opened 2019-02-19
lesson7 solution

aurel1212:
All the mailroom options should work. The mailroom main menu has an additional option (4) linked with a sub-menu that allows to add/delete/update. Also, the referential integrity should work: cannot delete a donor with a donation amount. First should delete all donations, and after that, delete donor.


ldconejo:
Hi Aurel,

For the first part, your implementation of the Department table is good. One thing to fix is that, when displaying Job data, the expectation is that you combine data from the Job and the Department table, so that you can display the name of the Department and not the code. Please submit a pull request once that change is done.

-Luis

PR #616, opened 2019-02-24
Lesson07: Part 1 & Part 2

Pjurek3:
added department to example Person/Job model.  Implemented  days with property which means it won't be stored in database but accessible in application.


Pjurek3:
Added part 2 with modifications to mailroom.  I removed functionality for challenge from functional programming exercise to simplify donation model.

ldconejo:
Hi Paul,

Part one looks great.

For part two, nice touch on using a setup file. Two of the assignment requirements is to add ways to update/delete data. This could be implemented as an option to update/delete donors and/or their donations. Please submit a new pull request once those fixes are made.

Also, when entering a new donation, the value of the donation is multiplied by a factor of 100. Not a big deal, I just found it interesting

Regards,

Luis

PR #618, opened 2019-02-25
lesson8 and lesson7 updated

aurel1212:
I updated the lesson7 database to show department names.
For lesson8 I had issues with the Mongo db encryption on my Mac. I have multiple Python installations and looks like I cannot configure the right Python version to work with the remote Mongo database. So, for Mongo db, I used the default mongo db database. For redis and neo4j  the remote connections work.


ldconejo:
Hi Aurel,

For assignment purposes, it's ok that you used the default Mongo database as alternative. What error did you get?

Regards,

Luis

PR #620, opened 2019-02-27
Ballard Lockes Assignment Complete

jeanruggiero:


students/jeanruggiero/class03/ballard_lockes.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def __enter__(self):
+        print("\nEntering {}".format(self))
+        self.stop_pumps()
+        self.open_doors()
+        self.close_doors()
+        self.start_pumps()
+        return self
+
+    def __exit__(self, type, value, traceback):
```
ldconejo: Please modify your exit method so that it handles the exception, printing out the error message, but not crashing the program (you will need __exit__() to return True for that).

jeanruggiero:
Done, thanks for the feedback!

https://github.com/UWPCE-PythonCert-ClassRepos/SP_Online_Course2_2018/pull/627

PR #625, opened 2019-03-01
Lesson 1 for Review

srepking:
I noticed that the generator test file functions test_intsum() and test_intsum2() are the same. I disregarded the first. Let me know if there was a different intention for these.

Thanks!


ldconejo:
Hi Shane,

Once more, welcome to the course!

Good job on comprehensions. For generators, you need to include test_generator.py, already tested to be passing, as part of your submission. Finally, for the iterator exercise, you need to figure out a way to get your iterator to behave as _range()_. Once those fixes have been incorporated, please do another pull request for review.

Thanks,

Luis

srepking:
Hi Luis,
I made some changes to the iterators, and added the generator test file.

Please let me know if you wanted IterateMe_1() to have more range
properties. I figured that the main purpose is to show that you need to
explicitly call __iter__ on a range object in order to get the iterator. I
change IterateMe_1() to also have this property, in that it is only an
iterator object if you call __iter__ on it. Otherwise, it is just an
iterable object.

I hope you understand what I mean. I fear I don't have all the lingo down
yet.

Thank you,
Shane

PR #626, opened 2019-03-01
lesson07 - part 1 - complete

githubtater:


students/roy_t/lesson07/populate_db_personjobdept.py
```diff
@@ -0,0 +1,166 @@
...
+    START_DATE = 1
+    END_DATE = 2
+    SALARY = 3
+    PERSON_EMPLOYED = 4
+    DEPARTMENT = 5
+
+    jobs = [
+        ('Reverse Software Engineer', '2011-02-28', '2012-04-24', 135000, 'Peter', 'IT'),
+        ('Database Normalizer', '2000-03-12', '2009-12-12', 99000, 'Peter', 'IT'),
+        ('Structural Engineer', '2015-12-09', '2019-02-02', 140000, 'Andrew1', 'SA'),
```
ldconejo: Note that you're entering the employee's name as _Andrew1, Andrew2, Andrew3_, which do not match any of the person names you have (just "Andrew" would be the closest).

students/roy_t/lesson07/populate_db_personjobdept.py
```diff
@@ -0,0 +1,166 @@
...
+
+    finally:
+        logger.info('database closes')
+        database.close()
+
+
+if __name__ == '__main__':
+    populate_people()
+    populate_departments()
+    # populate_jobs()
```
ldconejo: When _populate_jobs()_ is not commented out, it will cause the program to crash. If the existing database is deleted and new one created, not running _populate_jobs()_ will cause the _Jobs_ table to be empty. The fact that it crashes is probably related to the other issue with the _jobs_ variable, earlier in the code.

Your code needs to be able to populate all tables without a preexisting database file.

students/roy_t/lesson07/pprint_db.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    Print the job information that each person has held, including department info.
+    """
+
+    try:
+        database.connect()
+        database.execute_sql('PRAGMA foreign_keys = ON;')
+
+        for job in Job:
+            pprint.pprint(f'{job.person_employed} was employed as a {job.job_name} from {job.start_date} to {job.end_date}. A total of {job.duration} days.')
```
ldconejo: This is working fine, you still need to incorporate the name of the department when indicating the positions a person has held. For this, you will need to combine data from the Jobs and the Department tables (the Department table will return the name of the department from the acronym obtained in the jobs table (for example "SA" should return "Systems Architect" (or "Systems Architecture", which would make more sense as department name).

Note that in the Jobs table, you have the "IT" department, which doesn't exist in the Department table.

students/roy_t/lesson07/populate_db_personjobdept.py
```diff
@@ -0,0 +1,166 @@
...
+    Populate the db with department information.
+    """
+
+    logger.info('Populating departments...')
+
+    DEPT_NUM = 0
+    DEPT_NAME = 1
+    DEPT_MGR = 2
+
+    departments = [
```
ldconejo: Need to add the "IT" department to this one, since you have some job entries using it.

ldconejo:
Hi Roy,

See my comments on the code. Very easy fix on the issue that causes _populate_jobs()_ to crash. You also need to pull the Department's name as part of _pprint_db.py_ when showing job data. With those two, you'll get the full grade for the assignment.

Thanks,

Luis

PR #628, opened 2019-03-02
lesson 9 homework

aurel1212:


students/AurelP/lesson9/get_news_multithread.py
```diff
@@ -0,0 +1,178 @@
...
+    word = word.lower()
+    count = 0
+    for title in titles:
+        if word in title.lower():
+            count += 1
+    return count
+
+start_time = time.time()
+results = queue.Queue()
+sources = get_sources()
```
ldconejo: Try not to have any code executing on its own outside of the main execution block. Move these after line 72.

students/AurelP/lesson9/get_news_multithread.py
```diff
@@ -0,0 +1,178 @@
...
+        word_count += count_word(WORD, titles)
+    return (word_count, art_count)
+
+def worker(*args):
+    results.put(search(*args))
+
+if __name__ == "__main__":
+    thread_count = 3
+    temp = int(len(sources)/thread_count)
+    for i in range(thread_count):
```
ldconejo: Instead, iterate over the list of sources. To avoid running out of usage of the API, just iterate over a few source, for example, do:

```
for source in sources[:5]:
```

students/AurelP/lesson9/get_news_multithread.py
```diff
@@ -0,0 +1,178 @@
...
+        start_index = i * temp
+        end_index = start_index + temp
+        thread = threading.Thread(target  = worker, args = (sources[start_index:end_index]))
```
ldconejo: In _args_, pass one source and the name of the queue. The worker should put the results of the thread it handles into the queue.

students/AurelP/lesson9/get_news_multithread.py
```diff
@@ -0,0 +1,178 @@
...
+    art_count = 0
+    word_count = 0
+    for source in sources:
+        titles = get_articles(source)
+        art_count += len(titles)
+        word_count += count_word(WORD, titles)
+    return (word_count, art_count)
+
+def worker(*args):
+    results.put(search(*args))
```
ldconejo: You should pass a source and the queue to _worker_. Worker should still do a _.put()_ operation into the queue, but pushing in the resulting list from _get_articles(source)_. That will cause the queue to contain multiple lists of articles, one list returned by every thread.

students/AurelP/lesson9/get_news_multithread.py
ldconejo: After launching all your threads, remember to have a join() operation for each of the threads (outside of the _for_ loop that launched the threads), so that your program waits for all threads to complete before the word count begins.

ldconejo:
Hi Aurel,

I suggested a number of changes in the code that will help it take advantage of threading and correct a few issues that are present. 

At the end, you'll need to pop each group of articles and count occurrences of the word being searched for.

Let me know if you have any questions and submit a new pull request once all fixes are made.

Thanks,

Luis

PR #629, opened 2019-03-02
Lesson1 Ready for Another Review

srepking:


students/srepking/lesson01/iterator_1.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+    def __next__(self):
+        self.current += self.step
+        if self.current < self.stop:
+            return self.current
+        else:
+            raise StopIteration
+
+    def __iter__(self):
```
ldconejo: You'll need to modify ___iter__()_ to make _IterateMe2()_ behave as the _range()_ function.

students/srepking/lesson01/iterator_1.py
```diff
@@ -0,0 +1,81 @@
...
+        return self
+
+if __name__ == "__main__":
+
+    print("Testing the iterator")
+    for i in IterateMe_1():
+        print(i)
+
+    print("Testing again with start, stop, and step size set:")
+    it = iter(IterateMe_1(2,20,2)) #'it' is an iterable object. You need to call __iter__ to get the iterator. Just like
```
ldconejo: You don't need to for _it_ to an iterable object, instead, call your (modified to behave like range) IterateMe2:

```
it = IterateMe2(2,20,2)
```

This is how you could test / compare IterateMe2 and Range:

```
    it = IterateMe2(2,20,2) 
    for i in it:
        if i > 10:
            break
        print(i)

    print("Emulating range, so the cycle should start again after break:")
    for i in it:
        print(i)

    print("Now testing range")

    it = range(2, 20, 2)
    for i in it: 
        if i > 10:
            break
        print(i)

    print("The real range after break:")
    for i in it:
        print(i)
```

The output should look like this:

```
2
4
6
8
10
Emulating range, so the cycle should start again after break:
2
4
6
8
10
12
14
16
18
Now testing range
2
4
6
8
10
The real range after break:
2
4
6
8
10
12
14
16
18
```

ldconejo:
Hi Shane,

Generator test looks good. I made a number of comments on your code on what the iterator output should look like when modified to work as _range()_, including some code to compare the output with the built-in _range()_ function.

Regards,

Luis

srepking:
Thank you Luis,
Please take a look and let me know if this is what you had in mind. I
created a new Class called PsuedoRange that may be what you are looking
for? So now I have two variations on Classes that behave like range, in
that they would satisfy the test code you provided. However, perhaps I am
still not grasping intent.

Thank you,
Shane

PR #634, opened 2019-03-05
lesson4 and lesson9 reviewed

aurel1212:
In lesson9 I got confuse between number of thread list and queue. I believe I fixed.
For lesson 4, I did minimum changes to the original mailroom: I added decorators, and load/save options.


students/AurelP/lesson9/get_news_multithread.py
```diff
@@ -0,0 +1,84 @@
...
+    sources = get_sources() #[:5]
+    threads=[]
+    for source in sources:
+        thread = threading.Thread(target=worker, args=(source,results,))
+        threads.append(thread)
+        thread.start()
+        print("Thread %s started" % thread.name)
+    for thread in threads:
+        thread.join()
+    results = [results.get() for i in range(len(sources))]
```
ldconejo: This is currently not returning. You can try a simpler way to fetch matches and number of articles for your queue. One alternative is the following code:

```
    occurrences = 0
    articles = 0
    while not results.empty():
        matches = results.get()
        occurrences += matches[0]
        articles += matches[1]
    print("Word: {} found {} times in {} articles".format(WORD, occurrences, articles))
```

ldconejo:
Hi Aurel,

Lesson 4 looks good. For lesson 9, threading is now correctly implemented. Word count is not working, I suggested some code to go over the data collected in your queue, try that or something similar and submit a new pull request, should be an easy fix.

Regards,

Luis

PR #636, opened 2019-03-06
Fix to integration-test, added omitted files

rbillupsii:


ldconejo:
Hi Roy,

This one did the trick. Coverage is at 100% for both target files.

Thanks,

Luis

PR #646, opened 2019-03-09
Lesson 08

elnurad:
Submitting nosql files.
Mailroom is still work in progress. I keep getting an error when I run mailroom/src/mongodb.py
I'd like to make sure that the database is working properly so I created a query and tried to find it with find_one(). I get an error which says 'InsertManyResults' object has no attribute 'find_one' and I can't figure out what I am doing wrong. Any feedback would be very much appreciated!

Thanks.
Elnura


students/elnurad/lesson08/mailroom/src/mongodb.py
```diff
@@ -0,0 +1,46 @@
...
+          {'donor': 'John Clark',
+           'donations': 3000},
+          {'donor': 'Andrew Jones',
+           'donations': 8000}]
+
+def run_mongodb():
+    with login_database.login_mongodb_cloud() as client:
+        db = client['mailroom'] 
+        donations = db['donations']
+        donations = donations.insert_many(donors)
```
ldconejo: Hi Elnura,

The problem is here. You're reassigning your database object, _donations_, to the result of the _insert_many_ operation, losing the database object (which does have the _find_one_ method) in the process.

Try changing this line to:

```
results = donations.insert_many(donors)
```

ldconejo:
Hi Elnura,

See my comment on your code, the problem is that you're reassigning the _donations_ variable. Please feel free to submit additional pull requests if you run into other issues.

Regards,

Luis

elnurad:
Thank you, Luis!

PR #649, opened 2019-03-12
Lesson 9 submission.

astrawmyer:


students/astrawmyer/lesson09/multi_news.py
```diff
@@ -0,0 +1,75 @@
...
+threads = []
+for source in sources:
+    thread = threading.Thread(target=get_articles, args=(source,))
+    thread.start()
+    threads.append(thread)
+
+for thread in threads:
+    thread.join()
+
+for source in sources:
```
ldconejo: This block causes the entire article search to run single threaded, source by source, so no multi-threading is really being used. 

I would suggest you create a thread handler function that is called by _threading.Thread_. This function will take a queue and one of the sources and capture the titles returned by that source by pushing them into the queue.

Once all threads have been joined, have your code for word count go over each of the lists of titles that were pushed into the queue (this can be done sequentially, no need to multi-threading at this point) and return the word count.

ldconejo:
Hi Adam,

See my comments on your code. Code is currently running sequentially, tossing all multithreaded output.

-Luis

PR #652, opened 2019-03-13
Lesson 05 submittal

Percolator787:
Submittal for lesson 05:

- CoffeyDebugging.txt
- recursive.py
- simple.py
- syslogserver.py

### DIFF
--- a/students/Dennis_Coffey/lesson05/CoffeyDebugging.txt
+++ b/students/Dennis_Coffey/lesson05/CoffeyDebugging.txt
@@ -1,0 +1,100 @@
+What is wrong with our logic?
+There is no logic to check or return for the False condition
+
+Why doesn't the function stop calling itself?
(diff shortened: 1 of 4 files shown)


students/Dennis_Coffey/lesson05/CoffeyDebugging.txt
```diff
@@ -0,0 +1,100 @@
+What is wrong with our logic?
+There is no logic to check or return for the False condition
```
ldconejo: Hi Dennis,

The code that executes for the _False_ case (when n is not equal to 2) is _return my_fun(n/2)_, so that's not the problem. The issue is that if the initial value is not a multiple of 2 the loop will never end.

-Luis

ldconejo:
Hi Dennis,

For the logging assignment, your log file should reflect the current date, for example _2019_03_14.log_. Please submit a new pull request with a fix for this.

Regards,

Luis

PR #656, opened 2019-03-14
L09 resubmision with corrections

astrawmyer:
Noticed I did the search multithreaded and then did it again sequentially and extracted the titles from the sequential results in the original submission. This new version has everything sequentially and completes in 1 second.


ldconejo:
Hi Adam,

Thanks for implementing the necessary changes, submission looks good now!

-Luis

PR #660, opened 2019-03-16
Lesson 3 for Review

srepking:


students/srepking/lesson03/context.py
```diff
@@ -0,0 +1,52 @@
...
+              'Restarting the pumps.')
+        return self
+
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        #print('__exit__({}, {}, {})'.format(exc_type, exc_val, exc_tb))
+        if exc_type is ValueError:
+            print('Too many boats in Locke. Locke will not operate.')
+            return self
+        else:
```
ldconejo: This _else_ statement is unnecessary, as it is preceded by a _return_ statement (i.e., the rest of the code will only execute if the return statement does not execute, making the _else_ redundant.

PR #665, opened 2019-03-19
Lesson 08 Part 1 & 2

Pjurek3:
* completed part 1 of lesson 08
* did last part in mongo with some airplane data
* also did a lot of refactor on the mailroom for part 2 of lesson 08.  I missed part 1 which is why this is mixed in.  
* refactor is aimed at making mailroom decoupled from database backend.


Pjurek3:
added part 2.  Had a lot of trouble implementing Neo4j and this is just Mongo and Redis.

PR #668, opened 2019-03-20
tbrackney lesson 7

tbrack:


ldconejo:
Hi Travis,

For the Mailroom part, you need to add capabilities to add / delete data (could be donor and / or donation data) to get the full grade. I would recommend you try to finish quizzes 9 & 10, then assignment 10 (which allows you to select from a variety of final projects, memory profiler could be a good option, considering the amount of time left). After 10 is done, move on to the threaded downloader on lesson 9.

Regards,

Luis

PR #671, opened 2019-03-23
Lesson04 Submit

srepking:


students/srepking/lesson04/donors.py
```diff
@@ -0,0 +1,201 @@
...
+            self.to_json(to_file)
+
+    @staticmethod
+    def json_load():
+        """Create an instance of Group and load into this a list of donors
+        from a file written in json format."""
+        with open('json.txt', 'r') as from_file:
+            from_dict = from_file.read()
+            if from_dict == '':
+                return
```
ldconejo: Try to keep you _return_ statements within a method consistent. Your code is set to deal with the _None_ return here, but it's always a good practice for code calling your methods to rely on all possible returns being of the same type (dictionary, string, etc.).

ldconejo:
Hi Shane,

Good job on this one. For filenames, try to use .json as extension for your JSON files. I liked the fact that your code is able to deal with an empty JSON and recover by asking the user for donor data.

Finally, try to audit your code with Pylint or another linter, there are a few minor changes that could be made to improve readability and adherence to PEP8 practices.

Regards,

Luis

srepking:
Thank you for the quick review Luis! I will run code through pylint for
here on out.

srepking:
Hi Luis,
I am not sure I am following you here. Should I just return the actual
dictionary 'from_dict' and deal with it in the calling method instead of
trying to filter the results in this method?

Thank you!
Shane

PR #672, opened 2019-03-24
fixed sql queries on mailroom db as they were wrong and Lesson09 mult…

tlugosan:
…ithreading news scraper solution


students/tlugosan/Lesson09/threaded_news_finder.py
```diff
@@ -0,0 +1,87 @@
...
+        print(resp)
+        print(resp.text)
+        return []
+    data = resp.json()
+    titles = [str(art['title']) + str(art['description']) for art in data[
+        'articles']]
+    return titles
+
+
+sources = get_sources()
```
ldconejo: This line should have been within the main execution block, (after line 72). It works, but you should always avoid having these "rogue" lines of code hidden in-between functions, as it's easy to lose track of them, making debug difficult later on.

students/tlugosan/Lesson09/threaded_news_finder.py
```diff
@@ -0,0 +1,87 @@
...
+    for thread in threads:
+        print("joining thread {}", thread.name)
+        thread.join()
+    print("all threads finish")
+
+
+if __name__ == '__main__':
+    start = time.time()
+
+    news_homework(sources)
```
ldconejo: This line should be removed, _threading_example_ is the one calling _news_homework_, I assume this one here is from earlier in development.

ldconejo:
Assignment requirements are met, see my comments for improvements for future code. No need to do another pull request in this case.

Regards,

Luis

PR #673, opened 2019-03-25
Lesson 9

rbillupsii:


students/RoyC/lesson09/news_getter.py
```diff
@@ -0,0 +1,91 @@
...
+        print("ERROR: ", r.status_code)
+        return
+
+    data = r.json()
+    print("Getting sources..." )
+    for src in data['sources']:
+        sources.put(src['id'].strip())
+
+def get_articles():
+    while not sources.empty():
```
ldconejo: Interesting implementation. Most other students have had the thread launcher get one source and pass it to _get_articles_, this is the first time I see _sources_ implemented as a queue, which works well and simplifies the call to _get_articles_.

PR #679, opened 2019-03-30
Submitting for review per email

jeanruggiero:


students/jeanruggiero/class07/MailroomDB/mailroom.py
```diff
@@ -0,0 +1,158 @@
...
+        "Thank you so much for your generous donation of " + \
+        f"${amount:.2f}.\n\nWe really appreciate your donations " + \
+        f"totalling ${donor.donation_total:.2f}.\n" + \
+        "Sincerely, The Wookie Foundation"
+
+
+if __name__ == '__main__':
+
+    database = SqliteDatabase('mailroom.db', pragmas={'foreign_keys': 1})
+    database.connect()
```
ldconejo: These two lines are not necessary, you're already connecting to the DB on _mailroom_model.py_.

students/jeanruggiero/class07/MailroomDB/mailroom.py
```diff
@@ -0,0 +1,158 @@
...
+                break
+            except ValueError:
+                print('Please enter a number value for donation amount.')
+
+        print(); print(thank(name, amount))
+
+
+def create_report():
+    """Print a report of donors with a summary of their donation history."""
+    donor_report.create_report()
```
ldconejo: No need for a separate module on this one, you should simply do another _for donor in (Donor.select()):_ and print the different attributes you need in your report. I liked the fact that you created properties to get total number of donations and total amount donated directly in your Donor class, that's simplifies creating the report even more.

ldconejo:
Hi Jean,

Your code actually looks pretty good, regarding what's required for this assignment. I added a couple of suggestions on the report part, other than that, I would say it's almost ready. 

-Luis

students/jeanruggiero/class07/MailroomDB/mailroom_model.py
```diff
@@ -0,0 +1,81 @@
...
+        try:
+            query = (Donation
+                     .select()
+                     .where(Donation.donor == self.name))
+        except Exception:
+            print('Error')
+        return len(query)
+
+    @property
+    def donation_total(self):
```
ldconejo: You could have you query return the total value, instead of doing a _sum_ after the query:

```
    @property
    def donation_total(self):
        try:
            query = (Donation
                     .select(fn.SUM(Donation.amount))
                     .where(Donation.donor == self.name)).scalar()
        except Exception:
            print('Error')
        return query
```

PR #702, opened 2019-04-23
submit mailroom assignment for lesson 7

cocokaku:


ldconejo:
Hi Colleen,

As required in the assignment, please include functionality, preferably as part of _mailroom.py_ to perform update and delete operations on the database. It could be updating a donor's name or the value of a donation, the same goes for deletions, you could delete a donor or a particular donation.

Thanks,

Luis

PR #707, opened 2019-04-29
adding lesson5

jayjonesseattle:
Please grade lesson 5 assignments. debugging.text is the debugging assignment. simple.py is the logging exercise. As always, any feedback you may have is welcomed.  Thank you!

### DIFF
--- a/students/TerranceJ/lesson5/complicated_example.py
+++ b/students/TerranceJ/lesson5/complicated_example.py
@@ -1,0 +1,26 @@
+def main():
+	x = 'main'
+	one()
+
+def one():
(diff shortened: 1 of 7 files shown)


ldconejo:
Hi Jay,

For the logging assignment, everything looks good and the error is also reported by _syslogserver.py_ but the data is not being passed correctly, the error shows up like this:

127.0.0.1 :  <11>%(filename)s:%(lineno)-4d %(levelname)s %(message)s

Try to give it another try, it should be a relatively easy fix. Let me know if you have any questions.

-Luis

PR #709, opened 2019-05-03
Part 1 of Lesson 7

srepking:
Questions:
For this lesson, to find the total of days worked in a job, I did the date subtraction in python. Is it preferred to do this in Python or in SQLITE3? Could you let me know how to do this in sqlite instead of python? Thank you!

Also, please read the README for directions in creating and populating the database.


ldconejo:
Hi Shane,

One thing to correct is that the _Department_ table should not have a _days_held_ field, as several employees could work or have worked for the same department. That field belongs in the _Job_ table. 

On the other, the _Job_ table should have a foreign key containing the _dept_number_ of the department where the person was working for that position, when displaying the data, that _dept_number_ will be used to pull the department name for the _Department_ table (you're already doing something similar with your query, so this should be an easy change once the tables are modified).

Please correct the items above and submit a new pull request.

-Luis

ldconejo:
To your question, although this assignment requires you to have _total_days_ as a field, you could also calculate it on demand as you query the database. This is, sort of, what you would need to do (I'm repurposing your existing query):

```
        query = (Person
                 .select(Person, Job.job_name, Department.dept_name, (Job.end_date - Job.start_date).alias('total_days'))
                 .join(Job, JOIN.INNER)  # Joins person - > Job
                 .join(Department, JOIN.INNER))  # Joins Job -> Department

        query_tuple = []  # Create a list to hold person, job, and dept.
        for person in query:
            query_tuple.append((person.person_name, person.job.job_name,
                                person.job.department.dept_name, person.total_days))
        return query_tuple
```

This will subtract _Job.end_date_ and _Job.start_date_ and assign the result to the _total_days_ column, which you then see used as part of your _query_tuple.append()_ operation. The result will be incorrect, because the strings should be converted to date-time before the subtraction, but the general idea is the same: You can do some calculations on the go while running a query.

Hope this helps!

Luis

srepking:
Hi Luis,
I incorporated your comments and thought it was straight forward, but now I
keep getting this error during my query, 'Job' object has no attribute
'department'. Is this because the join of the Department table isn't taking
place? Thank you again!

# #######################################################
# Getting this error for this
# query. 'Job' object has no attribute 'department'
# #############################################3

query = (Job
         .select(Job, Department)
         .join(Department, JOIN.INNER))  # Joins Job -> Department

query_tuple = []  # Create a list to hold person, job, and dept.
for job in query:
    query_tuple.append((job.person_employed, job.job_name,
                        job.department.dept_name))

ldconejo:
Hi Shane,

Try this:

```
        query = (Job
                 .select(Job.person_employed, Job.job_name, Department.dept_name.alias("dept_name")) #, Department.dept_name)
                 .join(Department, JOIN.INNER).objects())  # Joins Job -> Department

        query_tuple = []  # Create a list to hold person, job, and dept.
        for job in query:
            query_tuple.append((job.person_employed, job.job_name,
                                job.dept_name))

        return query_tuple
```

The _objects()_ method causes all columns to be returned under the _Job_ table. Not the best explanation, but that's how it works. You can read more about it here:

http://docs.peewee-orm.com/en/latest/peewee/querying.html 

Let me know if this works.

Regards,

Luis

srepking:
This worked great. Thank you so much! I made a few changes to the code
shown above, and it all resolved nicely.

Thanks again,
Shane

PR #712, opened 2019-05-08
Submitting lesson 09

Percolator787:
Lesson 09 Submittal

- news_downloader.py

### DIFF
--- a/students/Dennis_Coffey/lesson09/news_downloader.py
+++ b/students/Dennis_Coffey/lesson09/news_downloader.py
@@ -1,0 +1,98 @@
+# -*- coding: utf-8 -*-
+"""
+News downloader using threading
+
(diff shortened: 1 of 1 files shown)


students/Dennis_Coffey/lesson09/news_downloader.py
```diff
@@ -0,0 +1,98 @@
...
+    if resp.status_code != 200:
+        print("Something went wrong with {source}".format(source))
+        print(resp)
+        print(resp.text)
+        return[]
+    data = resp.json()
+    # The url to the article itself is in the data['articles'][i]['url']
+    titles = [str(art['title']) + str(art['description'])
+              for art in data['articles']]
+    results.put(titles)
```
ldconejo: This works, but consider to have the queue be an input parameter, instead of just assuming that it exists as a global variable.

PR #715, opened 2019-05-09
Lesson07 Mailroom Ready for Review

srepking:


students/srepking/lesson07/mailroom/mailroom_sql.py
```diff
@@ -0,0 +1,163 @@
...
+logger = logging.getLogger(__name__)
+
+
+def create_database():
+
+    file_name = input('\nWhat would you like to name your new database?\n'
+                      'e - to exit\n')
+    if file_name == 'e':
+        return
+    else:
```
ldconejo: Remember no _else_ is needed after a _return_ statement, as the code following _else_ will only execute if the preceding _return_ statement does not.

students/srepking/lesson07/mailroom/mailroom_sql.py
```diff
@@ -0,0 +1,163 @@
...
+    for file in file_list:
+        logger.debug(f'Print file in file_list {file}')
+        if file.endswith('.db'):
+            db_file.append(file)
+            logger.debug(f'Print db_file,{db_file}')
+    if db_file is not None:
+        print(f"The following databases exist in current working directory,\n")
+        for file in db_file:
+            print(f"{file}\n")
+    if len(db_file) == 0:
```
ldconejo: Python will consider an empty string to be "False", so you could replace this one with:

_if not db_file_:

students/srepking/lesson07/mailroom/mailroom_sql.py
```diff
@@ -0,0 +1,163 @@
...
+    if file_name == 'e':
+        return
+    else:
+        if os.path.exists(file_name):
+            logging.info('Trying to delete the database.')
+            os.remove(file_name)
+            logger.info(f'Database is {file_name} has been deleted.')
+
+
+def more_choices():
```
ldconejo: Try to be consistent as to the different routes the flow of a function can follow. For example, if one of those routes returns an integer, all other routes should also return an integer (i.e., you have not scenario in which your function would not return at all or in which it would return a string or another data type). 

Having inconsistent returns will increase the likelihood of your code eventually running into situations in which a corner case causes your code to crash, on top of being comparatively difficult to debug.

ldconejo:
Good job on this one. See my comments for potential improvements, there's no need to submit a new pull request for this one.

-Luis

PR #716, opened 2019-05-09
Lesson 10 submittal

Percolator787:
Lesson 10 submittal exploring / demoing Python's memory profiler.  Profiled the functional programming mailroom example from last quarter to observe how much memory is used by each of the functions.  
  - mailroom_fp.py
  - mailroom_fp_memory_profiler.txt - Memory profile results

### DIFF
--- a/students/Dennis_Coffey/lesson10/mailroom_fp.py
+++ b/students/Dennis_Coffey/lesson10/mailroom_fp.py
@@ -1,0 +1,269 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+import datetime
+import os
+from copy import deepcopy
+from functools import reduce
+
+"""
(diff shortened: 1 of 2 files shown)


students/Dennis_Coffey/lesson10/mailroom_fp.py
```diff
@@ -0,0 +1,269 @@
...
+"""Update to mail room program to add decorator to certain functions to provide 
+   a line by line analysis of the memory profile of the function"""
+
+
+# Donor class - contains properties and methods for accessing and updating donor's data
+class Donor:
+    
+    def __init__(self, name, donations=None):
+        self.name = name
+        if donations == None:
```
ldconejo: This should instead by written:

_if donations is None_

PR #722, opened 2019-05-20
final commit for Lesson7

JeroenSweerts:


ldconejo:
Hi Jeroen,

For the first part of assignment 7 tables should more or less work like this:

- Department: Use a department code as primary key and also contain the name of the department's manager, as well as the department name. It should not hold the names of employees working at that department, as a single department will normally have multiple employees.

- Person: Should have a unique ID for each individual (some students use the person's name, which is ok for the purposes of this assignment, but not practical for the real-world) which works as primary key, it will also contain the person's name and other fields such as the town where they live.

- Job: Should have a person_ID and department_ID as foreign keys from the Person and Department tables, respectively.

Now the assignment also requires you to do a pretty print showing which person held which job at which department, which will require you to combine data from all tables in order to present the information to the user.

Please submit a new pull request once these changes have been implemented.

Regards,

Luis

PR #726, opened 2019-05-31
Lesson 09 Threaded Downloader

ldenney89:
Submitting Assignment, though I ran out of requests on the newsapi site, so I'm not 100% satisfied that all works without errors. 

File:
multithreading_news.py


ldconejo:
Hi Laura,

There are a few changes that you should make to get threading working correctly here.

1. To avoid running out of usage of the API, limit the number of sources you work with, at least while doing development. For example, just use 5 sources:

```
    sources = get_sources()
    sources = sources[:5]
```

2. Don't actively limit the number of threads, instead, assign one news source per thread. The _for_ loop in _thread_the_articles()_ will look like this:

```
    for source in sources:
        thread = threading.Thread(target=worker, args=(sources,))
        thread.start()
```

3. You need to join the threads once all of them are launched: Your code will not wait for any threads that are running when processing the title, so you need to join each thread before trying the process the data collected. This can be done easily with a _for_ loop, similar to the one used for launching the threads. 

-Luis

PR #732, opened 2019-06-20
Solution Lesson9

JeroenSweerts:


ldconejo:
Hi Jeroen,

The solution for lesson 5 looks good. Note that one of the requirements is that the file name for the log generated includes the corresponding date. Feel free to add a fix for this in order to get the full grade for this assignment.

Regards,

Luis

ldconejo:
For lesson 7, you also need to add the capability for the user to update data (for example, a donor's name or the value of an existing donation) and delete data (a donor and all his/her donations or specific donations from a donor).

PR #734, opened 2019-06-30
Made tests executable from outside directory

jaredmulholland:


jaredmulholland:
Lesson 9 and lesson 10 homework

jaredmulholland:
I am having issues with the json load and save (lines 187-201) in mail_room_class_3.py. When I run the code outside of the class object, it seems to work find and create a json object and load a json object. When I try to run the method as part of the class, I get the type error that 'Object of type Dict is not JSON serializable'.

ldconejo:
Hi Jared,

For the calculator assignment on lesson 6, there are still a few things to correct:

1.  When doing unit testing from outside of the _calculator_ directory, the run will error out because the _exceptions_ module being called by _calculator.py_. This can be fixed by adding a period (.) before _exceptions_ to tell Python to look for the file in the same directory as _calculator.py_:

`from .exceptions import InsufficientOperands`

2. Both Pylint and Flake8 indicate quite a large number of issues with your code. Run both and see if you can correct those issues:

$ python -m pylint calculator/
$ python -m flake8 calculator/

Regards,

Luis

ldconejo:
For lesson 9, your code is sending the full list of sources to each of the 10 threads you're launching, so different threads are actually calling the same source, which is not the idea. You also run out of API usage really fast, as the free tier only allow 500 requests every 12 hours. Also, your _titles_ variable is being overwritten by all threads, so the end result you see is what the last thread to finish wrote on it.

One way to do it is as follows:

1. Have your main flow send only one source to each thread worker. For debug purposes, use only the first 5 or 10 news sources in _sources_ (similar to your current limitation of launching only 10 threads). Your could do:

```
for source in sources[:5]:
    # launch each thread here
```

2. Each thread uses its source and then calls _get_articles_ which returns a list of news titles.
3. Push this list of titles into a _queue_ (you're currently importing the module but not using it). Queues are thread-friendly, which means you can have multiple threads pushing their own lists of titles into the same queue and Python will only allow one thread at a time to write into the queue, which is very convenient. In the end, the queue will be a collection of lists of titles.
4. After you join all your threads (confirming that they all finished) then do your word count by popping each list of titles out of the queue and counting occurrences of your target word within that list. Do the same for the next list until the queue is empty.

ldconejo:
Also, don't name your script _threading.py_ and doing that effectively overrides the _threading_ module , preventing your code from running.

PR #735, opened 2019-07-17
added code to file

jaredmulholland:
thanks - I must have accidentally deleted the code before committing previously.


jaredmulholland:
renamed threading.py to news_api.py. I ran the code 5 times (got an error on the 5th time due to the 250 requests per 12 hours).

jaredmulholland:
Thanks for the priority list. I was halfway through 7 problem 1, so I finished that up first.

ldconejo:
Excellent. I just reviewed your implementation of lesson 7, part 1. I'm glad you worked on that one, I thought it was going to be tough to finish with the amount of time available, but definitely knowing how to implemented a SQL database on Python is very useful. Your grade is now at 72, keep up the good work, just one last effort!

Regards,

Luis

jaredmulholland:
Yes, that one was a little easier with me because I work with relational
databases at work quite a bit.

I just created a pull request for the calc files.

Curious if you had any thoughts on my lesson 4 homework. I'm having trouble
integrating the json methods into my class (they seem to create a file just
fine out of the class).

Thanks for your help
Jared

jaredmulholland:
Luis,

I'm trying to finish up this weekend so I can pass. I just pushed the water
regulator files and then created a pull request. I'm having issues with the
unit_test.py (I was having all sorts of issues with the decider file, but I
think I have that figured out now).

For the unit_test.py, when I try to run it, I get the error: Attribute
Error: 'DeciderTests' object has no attribute 'decider'

I think this has something to do with how I'm reading the modules in, but
I'm not sure. I tried creating a Decider instance separately and that
worked, including running the method decide. Anyways, I'm stuck, and I'm
trying to get this done before my time runs out :).

Thanks,
Jared

PR #737, opened 2019-07-20
fixed typo

jaredmulholland:
had typo in line 35, 36


jaredmulholland:
The issue seems to be:
1. load json - seems to load just fine
2. save json - doesn't seem to save the new donors that I add
3. it also seems like my create report doesn't work anymore.

jaredmulholland:
made pylint and flake8 changes

ldconejo:
Hi Jared,

I'm going to keep this pull request open for now, so that we can debug this together. Some things to change for lesson 4:

1. Try to avoid having any code running outside of the _main_ block. You currently have all the default donor / donations database initialization somewhere at the beginning of your code. This is not a good idea, as it's easy to forget that you have some one-liner, somewhere in between the declaration of two clases.

2. Within _Donor_ class, _add_donation()_ should NOT be a property, it's a class method with only one job: Adding new donations to a _Donor_ instance.

3. Also within _Donor_class_, the _thank_you()_ method should not be a property either.

4. When creating a new donor, the way you're coding it, it will turn what you're passing as donations into a list. That is, if donations == 1000, it will turn it into [1000]. If it is [50000, 65000] it will become [[50000, 65000]] --> a list within a list. My recommendation is to change the _Donor_ init not to to convert _donations_ to a list:

```
        if donations:
            self.donations_js = donations # This was [donations]
        else:
            self.donations_js = []
```

Note that you will need to change the calls to _add_donor()_  to reflect the fact that they're no longer class properties. Same applies for any properties that I'm recommending you to change back to class methods.

After doing these changes, _create_report()_ starts working:

```
$ python mail_room_class_3.py 

Main Menu:
1. Send a Thank You
2. Create a Report 
3. Send Letters
4. Create Projection 
5. Save JSON
6. Load JSON
7. Quit
Please Choose an Option: 2


Donor Name           |    Total Given|   Num Gifts |   Average Gift
Jared Mulholland     $      10,000.00            1 $      10,000.00
Chris Cornell        $     115,000.00            2 $      57,500.00
Ben Shepard          $      52,000.00            2 $      26,000.00
Kim Thayil           $      83,000.00            3 $      27,666.67
```

The next step would be to get _send_thank_you()_ to work, which should be relatively straightforward. With this basic framework working, go back to saving/loading JSON files. 

You're actually quite close to having this one ready, keep up the good work!

-LUis

jaredmulholland:
I'm having trouble with this one. I think I made the add_donation and add_donor methods class methods, I took out the setter, but it doesn't seem to be working.

jaredmulholland:
Hoping to get some partial credit for assignment 2 in lesson 7

jaredmulholland:
Still can't get tests to run due to issues with reading the modules

jaredmulholland:
Hi , I submitted assignments 6b and 7b, hoping to get some partial credit.
I’m still not getting 4 to work. Wondering if I should submit as is
depending how many points I need to get to 80.

Thanks,
Jared

ldconejo:
Hi Jared,

For assignment 7, part 2, the implementation is good for 50% of the grade, you could try adding some basic functionality for adding, updating and deleting data from the database to get the full grade.

For assignment 6, part 2, unit and integration tests still have too many issues.

For assignment 4:

- Line 358: Change the donations from the first user to be a list:

    jared = Donor("Jared Mulholland", [10000])

- Line 106: _add_donor()_ should not be a class method, you're creating a instance of _DonorGroup_. The method should look like this:

```
    def add_donor(self, new_donor):
        self.donor_list_js.append(new_donor)
        self.donor_dict_js[new_donor.fullname_js] = new_donor.donations_js
        return self.donor_list_js

```
- Lines 364-366: Convert to method calls:

```
    dg.add_donor(chris)
    dg.add_donor(ben)
    dg.add_donor(kim)
```

- Line 188, _save_donors_json()_ should not be a class method. The save method should look like this:

```
    def save_donors_json(self):
        donors_json = json.dumps(self.donor_dict_js)
        with open('donor_db.json', 'w') as donor_file:
            donor_file.write(donors_json)
            print('JSON file created')
```
You still need to add some code in _load_json()_ to load the data into the actual database, so that running _create_report()_ will show the data loaded from the JSON file.

Try to finish assignment 4, hopefully tonight and I will review it right away.

Regards,

Luis

jaredmulholland:
Thanks - I had tried to change the add_donor method to look like that
before, but got the message:

An attribute defined in mail_room_class_3 line 164 hides this method
pylint(method-hidden)
Do I need to worry about this?

ldconejo:
Don't worry about that for now, let's try to get the save/load JSON functionality enabled, which is the main objective of the assignment.

jaredmulholland:
Thanks, I incorporated the changes you suggested

